PR #161, opened 2014-12-22
Default AssemblyInfo

varon:
This allows tests to be built and run from within an IDE. 

The build scripts using `rake` were tested (on windows) and confirmed to be working.

### DIFF
--- a/Tests/App.config
+++ b/Tests/App.config
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8" ?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <runtime>
     <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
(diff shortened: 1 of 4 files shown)


haf:
@mausch we've forked Fuchu until we either get access to push to Fuchu nuget proper or you release the same updates we've done (maintenance release) https://www.nuget.org/packages/Fuchu-suave/

haf:
This is not the solution. This is the code that's failing:

```
let version =
  Assembly
    .GetExecutingAssembly()
    .GetCustomAttributes( typeof<AssemblyVersionAttribute>, true )
    |> Seq.head
    :?> AssemblyVersionAttribute
    |> fun a -> a.Version
```

It means we have to have a default AssemblyVersion or this code will fail. Let's change the code instead.

ademar:
Cool! I was thinking all the time the bug was in Fuchu.

mausch:
@haf What "updated dependencies" do you need from Fuchu? My package doesn't even have any listed dependencies: https://www.nuget.org/packages/Fuchu
I only have one pull request on my repo and it's about filtering, could you please send a pull request or open an issue describing what you need?

haf:
@mausch you have an implicit dependency on FSharp.Core, also, for FsCheck, PerfUtil there are new versions available... I can open issues for them

PR #176, opened 2015-01-03
[BREAKING] string -> Uri, ip+port -> SocketBinding

### LINKED ISSUE #143: HttpRequest.url missing scheme, host, port etc
It would be good to have the url properly passed along, rather than just getting the `path` value of it in `url` of `HttpRequest`. This is needed for Hawk authentication.

haf:
Partially fixes #143 

This should not entail much breakage for the end-consumer of the library.

### DIFF
--- a/src/Suave/Tcp.fs
+++ b/src/Suave/Tcp.fs
@@ -9,18 +9,18 @@ open Suave.Logging
 open Suave.Sockets
 
 /// The max backlog of number of requests
-let MAX_BACK_LOG = Int32.MaxValue
+[<Literal>]
+let MaxBacklog = Int32.MaxValue
 
 type StartedData =
   { start_called_utc : DateTimeOffset
     socket_bound_utc : DateTimeOffset option
-    source_ip        : IPAddress
-    source_port      : uint16 }
+    binding          : SocketBinding }
 with
   override x.ToString() =
     sprintf "%.3f ms with binding %O:%d"
       ((x.socket_bound_utc |> Option.fold (fun _ t -> t) x.start_called_utc) - x.start_called_utc).TotalMilliseconds
-      x.source_ip x.source_port
+      x.binding.ip x.binding.port
 
 
 /// Disconnect a socket for reuse
@@ -89,7 +89,6 @@ let create_pools logger max_ops buffer_size =
 // i.e: export MONO_GC_PARAMS=nursery-size=128m
 // The nursery size must be a power of two in bytes
 
-/// Argh!! @ System.Net.Sockets.SocketException: Address already in use
 let private a_few_times f =
   let s ms = System.Threading.Thread.Sleep (ms : int)
   let rec run = function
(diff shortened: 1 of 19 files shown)


haf:
Ready to merge.

haf:
Failing with error:

```
Expected: "http://[::1]:8084/a?b=2"
Actual: "http://[0000:0000:0000:0000:0000:0000:0000:0001]:8084/a?b=2"
(00:00:00.0419709)

HttpBinding/IPv6 uri 2: Failed:
uri
Expected: "http://[::1]/a?b=2"
Actual: "http://[0000:0000:0000:0000:0000:0000:0000:0001]/a?b=2"
```

but passes on my own mono installation on OS X

src/Suave/Tcp.fs
```diff
@@ -102,26 +101,23 @@ let private a_few_times f =
...
-let tcp_ip_server (source_ip : IPAddress,
-                   source_port : uint16,
-                   buffer_size : int,
+let tcp_ip_server (buffer_size        : int,
                    max_concurrent_ops : int)
-                  (logger : Logger)
-                  (serve_client : TcpWorker<unit>) =
+                  (logger             : Logger)
+                  (serve_client       : TcpWorker<unit>)
+                  (binding            : SocketBinding) =
```
haf: Reordered to allow partial application.

src/Suave/Tcp.fs
```diff
@@ -9,18 +9,18 @@ open Suave.Logging
 open Suave.Sockets
 
 /// The max backlog of number of requests
-let MAX_BACK_LOG = Int32.MaxValue
+[<Literal>]
+let MaxBacklog = Int32.MaxValue
```
haf: Follow name standard for constants

src/Suave/Types.fs
```diff
@@ -644,6 +655,14 @@ type SuaveConfig =
 [<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
 module SuaveConfig =
 
+  let to_runtime config content_folder compression_folder =
```
haf: Since it's a lotta reading from the config...

src/Suave/Web.fs
```diff
@@ -278,20 +277,37 @@ module ParsingAndControl =
 
   /// Process the request, reading as it goes from the incoming 'stream', yielding a HttpRequest
   /// when done
-  let process_request is_secure (ctx : HttpContext) : SocketOp<HttpContext option> = socket {
+  let process_request ({ connection = connection
+                         runtime    = runtime } as ctx)
+                      : SocketOp<HttpContext option> = socket {
 
-    let! (first_line : string), connection = read_line ctx.connection
+    let! (first_line : string), connection' = read_line connection
```
haf: Making new values explicit instead of shadowing

src/Suave/Web.fs
```diff
@@ -278,20 +277,37 @@ module ParsingAndControl =
...
 
-    let request =
-      HttpRequest.mk http_version url meth headers raw_query
-        (parse_trace_headers headers) is_secure connection.ipaddr
+    // TODO: if client Host header not present, respond with 400 Bad Request as
+    // per http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html
+    let host =
+      headers %% "host" |> Option.get
+      |> function
+      | s when System.Text.RegularExpressions.Regex.IsMatch(s, ":\d+$") ->
```
haf: Would be nice with real parsers like Freya has, for all of this stuff, so that not both the regex and then the substring has to match.

src/Suave/Web.fs
```diff
@@ -278,20 +277,37 @@ module ParsingAndControl =
...
 
-    return Some { ctx with request = request; connection = connection }
+    let request =
+      HttpRequest.mk http_version
+                     (runtime.matched_binding.uri path raw_query)
+                     (ClientOnly host)
+                     meth headers raw_query
+                     (parse_trace_headers headers)
+                     runtime.matched_binding.scheme.secure
+                     connection''.ipaddr
```
haf: Should we make it explicit it's the client's address?

src/Suave/Web.fs
```diff
@@ -402,7 +418,7 @@ module ParsingAndControl =
       let verbosef = Log.verbosef runtime.logger "Suave.Web.request_loop.loop" TraceHeader.empty
 
       verbose "-> processor"
-      let! result = process_request (match runtime.protocol with HTTP -> false | HTTPS _ -> true) ctx
+      let! result = process_request ctx
```
haf: Simplify, yes, plx

Issue #83, opened 2014-04-19
[web] Read suave configuration from file


haf:
As a user of puppet,
When deploying a service,
I need to configure location of temporary writes, for writeability and security,
And that configuration needs either to be command-line based or configuration file based.

MatthiasHarvey:
I'll try implement this. I'll use JSON + FsPickler unless there are any other suggestions.

haf:
Actually, from reading FsPickler's docs, its JSON format is mostly geared towards being isomorphic to the CLR constructs - and being able to represent all sorts of types, cyclic dependencies etc. I suggest you instead to explore one of these paths:

 - using paket to source-reference the Json files of this folder https://github.com/fsharp/FSharp.Data/tree/master/src/Json - no dependencies, compiled into solution, versioning
 - using Newtonsoft.Json in a separate .Config assembly
 - using Newtonsoft.Json in the same assembly and il-merging/il-packing it with https://github.com/gluck/il-repack or ilmerge http://www.microsoft.com/en-us/download/details.aspx?id=17630 (I've had some problems getting both of [those](https://github.com/gluck/il-repack/issues/68) [to run](https://github.com/gluck/il-repack/issues/66) [on mono](https://github.com/logary/logary/pull/50).

Here's [some background discussion on serialisation with json](https://github.com/SuaveIO/suave/issues/89#issuecomment-40884979). I've investigated the JSON stuff in the BLC, but [it has its own problems](https://twitter.com/henrikfeldt/status/539405194485039104), plus and as for Fleece, it has unwanted dependencies.

So that's some of the research I've done the last year about JSON on .Net... I would like to hear your opinion on what path to take from here, but I don't think FsPickler is right for human-readable JSON.

haf:
All that said; I have one reservation to it. Since we already depend on FsPickler, iff FsPickler can read and generate human-readable JSON, that's probably a good way forward - and a better one than my own alternatives.

ademar:
Or `DataContractJsonSerializer` in `System.Runtime.Serialization.Json` ...

ademar:
FsPickler will be most probably the right aproach here so lets use that.

haf:
As I wrote in the tweet linked, I really don't think System.Runtime.Serialization.Json is the right alternative.

ademar:
We have agreed on FsPickler since we already depend on it. I don't see how the tweet is relevant thought.

MatthiasHarvey:
[...]

PR #187, opened 2015-01-07
Read suave configuration from file (#83)

MatthiasHarvey:
Used Newtonsoft.Json as external dependency. Had to do some fiddling to get it to build. Added FsPickler.Json to suave/Suave/paket.references and suave/.paket/paket.dependencies then ran paket install.


ademar:
Hum .. I don't like changing `mime_types_map` type .. this is going to break stuff

Can we create an intermediary type to perform the JSON conversion ?

ademar:
Suave/app.config does nothing AFAIK should not be there

ademar:
actually I'm OK with the `mime_types_map` change

src/Suave/Http.fs
```diff
@@ -99,22 +99,24 @@ module Http =
...
-      | ".jpe"
-      | ".jpeg"
-      | ".jpg" -> mk_mime_type "image/jpeg" false
-      | ".js"  -> mk_mime_type "application/x-javascript" true
-      | ".exe" -> mk_mime_type "application/exe" false
-      | ".txt" -> mk_mime_type "text/plain" true
-      | _      -> None
+    let default_mime_types_map =
+      let types = [
+        (".bmp" , MimeType.mk "image/bmp" false);
```
ademar: i think you can safely remove parentheses and semicolons here

MatthiasHarvey:
It wasn't building without Suave/app.config previously but it is now

MatthiasHarvey:
Should be fixed now

src/Suave/Types.fs
```diff
@@ -355,7 +355,7 @@ type Host =
   /// The client's Host header is this value
   | ClientOnly of string
   /// The
-  | Forwarded of forwarded_for:string * Host
+  | Forwarded of string * Host
```
haf: ehm, no?

src/Suave/Suave.fsproj
```diff
@@ -196,4 +196,89 @@
       </ItemGroup>
     </When>
   </Choose>
+  <Choose>
```
haf: I will revert this if it gets merged, it's massively breaking and a pain in the ass to work with; like I've said three times so far, if you want to use Newtonsoft.Json then either ILMerge it or make it a stand-alone lib.

src/Suave/Web.fs
```diff
@@ -526,14 +526,18 @@ let web_server (config : SuaveConfig) (webpart : WebPart) =
...
-    listen_timeout   = TimeSpan.FromSeconds(2.)
     ct               = Async.DefaultCancellationToken
-    buffer_size      = 8192 // 8 KiB
-    max_ops          = 100
-    mime_types_map   = Http.Writers.default_mime_types_map
-    home_folder      = None
-    compressed_files_folder = None
-    logger           = Loggers.sane_defaults_for LogLevel.Info }
+    logger           = Loggers.sane_defaults_for LogLevel.Info
+    properties       =
```
haf: Try to keep with the code standard in the readme and the rest of the code-base

src/Suave/Types.fs
```diff
@@ -778,50 +778,46 @@ let request f (a : HttpContext) = f a.request a
...
+
+open System.Threading
+open Nessos.FsPickler.Json
+
+/// The core configuration of suave. See also Suave.Web.default_config which
+/// you can use to bootstrap the configuration:
+/// <code>{ default_config with bindings = [ ... ] }</code>
+type SuaveConfig =
+  { /// Static configuration variables
+    properties              : ServerProperties
```
haf: What do you say about naming this `props` instead? Less typing.

src/Suave/Suave.fsproj
haf: The underlying reason is that it will cause assembly binding conflicts, because of the frequency that Newtonsoft.Json is being used. Also, it's very backwards-incompatible for consumers who don't have the latest newtonsoft.

src/Suave.Tests/Proxy.fs
```diff
@@ -37,9 +37,8 @@ let proxy =
 
   // let sslCert = X509Certificate.FromPKCS12(BIO.File("suave.p12","r"), "easy")
   // let proxy_config = { default_config with bindings = [ HttpBinding.Create(Protocol.HTTPS(sslCert), "127.0.0.1", 8084) ] }
-  let proxy_config =
-    { default_config with
-        bindings = [ HttpBinding.mk HTTP IPAddress.Loopback 8084us ] }
+  let customProperties = { default_config.properties with bindings = [ HttpBinding.mk HTTP IPAddress.Loopback 8084us ] }
```
haf: Can we try to keep a line length of 80 chars max?

examples/Load/Program.fs
```diff
@@ -11,12 +11,16 @@ open Suave.Logging
...
-  { default_config with 
-      bindings = [ HttpBinding.mk' HTTP "127.0.0.1" 8082 ]
-      ; buffer_size = 2048
-      ; max_ops = 10000
-      ; logger = logger }
+let config : SuaveConfig =
+  let custom_properties =
+    { default_config.properties with
+        bindings = [ HttpBinding.mk' HTTP "127.0.0.1" 8082 ]
+        ; buffer_size = 2048
```
haf: no `;`

examples/Example/Program.fs
```diff
@@ -53,8 +53,8 @@ let sleep milliseconds message: WebPart =
 
 // Adds a new mime type to the default map
 let mime_types =
-  Writers.default_mime_types_map
-    >=> (function | ".avi" -> Writers.mk_mime_type "video/avi" false | _ -> None)
+  Writers.default_mime_types_map.Add
```
haf: Use `Map.add`?

src/Suave/Types.fs
MatthiasHarvey: I get these errors if I revert it: "Unexpected symbol ':' in member definition" and "The type 'forwarded_for' is not defined". What do you suggest I do?

src/Suave/Suave.fsproj
MatthiasHarvey: Sorry, I'm new to all of this. Does making it a stand-alone lib mean that I will put the json parsing code into a separate dll?

haf: Yes exactly. No problem :). I've done something similar in Logary; it's really about taking a JSON and creating a DTO.

You're going good job so far, btw. =)

MatthiasHarvey: Ok cool. Thanks :)

MatthiasHarvey:
I'm fixing the ServerPropertiesJson thing now

haf:
The icing on the cake would be to write a unit test for the functionality (or integration test, depending on what you call it).

MatthiasHarvey:
Ok, I'll try to write a test.

haf:
Just reading a 'real' file is fine. There are helpers here https://github.com/SuaveIO/suave/blob/master/src/Suave.Tests/TestUtilities.fs#L27

MatthiasHarvey:
Alright, thanks

MatthiasHarvey:
Is a separate package for JSON going to be implemented as mentioned here #191? I'm having some trouble serializing 'bindings : HttpBinding list' so it might be better if I use FsPickler/Newtonsoft.Json again instead of trying to get around this issue.

ademar:
Since this is going in Suave.dll I would like to have no dependencies here. You could work around it by introducing an intermediate type that is easy to serialize with the BCL serializer.

MatthiasHarvey:
I was thinking about having the config reader in the Json package. Could that work? Otherwise I'll try with the intermediate type.

ademar:
> I was thinking about having the config reader in the Json package. Could that work? 

No, I think we should include this functionality in Suave.dll because it can be implemented without dependencies. We want to maximize utilization.

MatthiasHarvey:
Ah ok. I'll do it like you said then.

MatthiasHarvey:
It's mostly working, there's just one problem. If I try do
```f#  
scheme = if old_bind.scheme = "http" then HTTP else HTTPS
```
then I get the error "This expression was expected to have type
 Protocol 
but here has type
 ITlsProvider -> Protocol D:\suave\src\Suave\Types.fs"

ademar:
``fsharp
type Protocol =
  | HTTP
  | HTTPS of ITlsProvider
```

**Comment by ademar** [comment]
Example:
```fsharp
open Suave.OpenSsl.Provider

let ssl_cert = new X509Certificate("suave.pfx","easy");
let cfg =
  { default_config with
      bindings =
        [ HttpBinding.mk HTTP IPAddress.Loopback 80us
          HttpBinding.mk (HTTPS (open_ssl ssl_cert)) (IPAddress.Parse "0.0.0.0") 443us
        ]
      timeout = TimeSpan.FromMilliseconds 3000. }
```

ademar:
I'm not sure how can we specify the TLS provider in the config json ... cc: @haf

haf:
I would have the config take a string of a pfx or pem file with optional certificate chain file, and then after reading the config structure, if it has that string, create the F# values representing the files.

ademar:
Currently the TLS providers is packed in the sum type `HTTPS tls_provider`, we need to introduce it then in a different way .. perhaps making it part of the runtime record.

haf:
Closing due to the lack of activity.

ademar:
I''ll try to do something with this.

PR #188, opened 2015-01-10
add parsing decimals to Model module

theimowski:


haf:
@theimowski thanks! =)

Issue #153, opened 2014-12-05
[discussion] Freya + Suave?


haf:
I posit that we start combining the two projects; for Suave it will mean OWIN support, increased [RFC](https://github.com/freya-fs/freya/tree/develop/src/Freya.Typed/Types) compliance, [increased accuracy](https://github.com/freya-fs/freya/blob/develop/src/Freya.Typed/Lenses.fs) in dealing with modifications to HttpResult and support for the 'http application protocol', [like linked here](https://github.com/SuaveIO/resources/tree/master/routing).

We could start by finishing the OWIN support as that might make it much easier to use the monadic abstractions that Freya provides on top. Especially during the integration stage. [Freya's license](https://github.com/freya-fs/freya/blob/master/LICENSE.txt) supports remixing.

I'd like this thread to be the discussion about how we can go about improving Suave with the help of Freya's code and people to make the best F# web server there is! /cc @panesofglass @ademar @kolektiv

### Suggested List of Action Points

 - [x] Make it so that all modules-per-type have Lens signatures #181 
 - [x] Finish OWIN support to let Suave be a server #142 
 - [x] Create a PoC/spike on passing data through a pipeline like Freya does it, to see if there's overlap
 - [x] Consider what can be done to make the parsers more sturdy?

haf:
@kolektiv - you said previously you don't think the applicatives of WebPart would convey enough information about their values - your pipeline -- but can't we have the same abstraction on top of them as we have in OWIN, as we have the `user_state` to possible pass values in?

haf:
Added some action points.

kolektiv:
@haf although I've wrapped our (Freya) monadic state in an extra record to carry some information around, I don't think there's anything which would stop it being modelled as applicatives. Thinking about it, I started with a monadic assumption with Freya, as I started with a computation expression centric approach, and that seemed a reasonable assumption. As everything fundamentally builds on that computation expression type, I've not seen a reason to change it, but I am certainly interested in discussing it further - it's something that came up with @panesofglass too recently.

The CE syntax in Freya I think works well, but is monadic the right assumption everywhere? I'm certainly open to changes in thinking :)

panesofglass:
Linking in #97. Have you considered migrating the signature to the `AsyncState` approach of `HttpContext -> Async<'T * HttpContext>`? That would provide an easier transition to Freya, I think.

panesofglass:
[...]

PR #191, opened 2015-01-14
Json module now: 1) uses Newtonsoft 2) has error handling 3) includes a map_json_async function

jamesjrg:


haf:
@jamesjrg great =) Thanks

ademar:
sorry I'm reverting this.

ademar:
henrik .. i was writing  a large response when you merged this.

haf:
Sorry

ademar:
my complaints about this PR. 

- it includes changes to files unrelated to the issue at hand.
- it 'solves' two issues
- and most important - we haven't agreed upon using Json.NET
- and last we haven't agree to wrap in a try-catch

the last two issues you have personally pronounced against in the past.

I think we should discuss a little bit this things

jamesjrg:
- The changes to Suave.Testing was to make it easier to write a test of the function in the issue at hand, so is related

- The change to include a different Microsoft.FSharp.Targets is necessary to make it build with Visual Studio at all, because as far as I can tell it relies on F# 3.1. You could use an App.config binding redirect instead, but this way is more explciit.

- I agree noone has agreed on using Json.NET, if this is a bad idea for some reason then of course this can't be merged. My main problem with DataContractJsonSerializer  is that it requires you to put DataContract attributes all over your classes

- Maybe the try/catch part could be made optional somehow? It seems nice to at least have the option of a sensible default error response in the case of a JSON parse error, but I can see that this might not always be what people want

ademar:
Json serialization have been source of most discord see for example https://github.com/SuaveIO/suave/issues/83 and there are others.

I like the BCL serializer because no dependencies and it have worked so far for me. Henrik suggested in that thread that if Json.NET were to be used it will "need to ILMerge it or make it an external lib". Ultimately I think I wouldn't mind using Json.NET, it is true that is anoying decorating the fields.

Regarding `map_json` I'm still inclined to leave it in its minimal form with no try-catch, if the programmer wishes to send a 400 he can wrap it.

jamesjrg:
If you do decide to internalise the library some other options are:

1. Using LibZ instead of ILMerge. According to their website the author of ILMerge said ""As the author of ILMerge, I think this is fantastic! If I had known about this, I never would have written ILMerge"

2. Perhaps copy/pasting the Simple.Json files into your project, because it is smaller and simpler to include in the dll than Newtonsoft, However it doesn't seem to be actively maintained, so maybe in the long run that's not a good idea. (NB SimpleJson without a period is a different project altogether!)

As has been apparently been discussed before, the BCL serializer has a long list of other minor problems, e.g. it serializes dictionaries to a list of key/value pairs, which isn't what most people would really want when returning a dict as JSON.

Regarding try/catch, I can see that it might be best to leave it up to client code, but a few thoughts:

1) If it does use an external library, the client code will need to reference that library to be able to specify the correct type of exception to catch, making all client code dependent on the inner workings of Suave's JSON functionality.

2) I find it hard to reason about what happens in the interaction of WebPart composition, async, and exception stack unravelling. Can you put try/catch at any point within WebPart composition, or are there some combinations that are not allowed? Are there any surprising edge cases?

3) Most exceptions will be AggregateExceptions, but if you try/catch around map_json the exception will be direct, not inside an aggregate, so you will have to catch different sorts of exceptions depending on where precisely where your try/catch is placed.

jamesjrg:
There is also SimpleJson (without the period), which has the whole thing in a single file: https://github.com/facebook-csharp-sdk/simple-json/blob/master/src/SimpleJson/SimpleJson.cs

However it is C# so adding it as a single file in the project would not be as simple as it might be, and unlike Json.Net and Simple.Json it has no support for F#-specific types. It also doesn't look like it is maintained any more.

ademar:
I think the best would be to move the JSON code to its own assembly Suave.Json.dll and nuget it. This way we can take the dependency on Newtonsoft.Json.  Does this work for you @haf ?

haf:
@ademar That's a great solution. I'd be +1 on it, because it's then very obvious why one would take that dependency.

@jamesjrg It *is* hard to reason about exceptions. In general, in suave, web parts are "sequentially" executed and if one of them throws an exception, that exception is directly caught by suave and given to the exception handler in the config. There's really not much more than that to it.

If you want to catch exceptions inside your functors/web parts you'll have to wrap them yourself in an exception handler. `GET >>= (handle_exns save_user_details) >>= Writers.set_header "Content-Type" "application/json"`

I normally reason with async with -- it's like normal sequential code, but it gobbles down my call-stack, so I have to write code that doesn't throw exceptions, or it will be painful. That's why I wrote logibit.hawk -- because hawknet is so utterly stuck with exceptions that it becomes hard to work with it.

jamesjrg:
I agree that exceptions are generally difficult to work with, the best thing would be if JSON libraries offered TrySerialize/TryDeserialize methods that you could then pattern match on, but unfortunately they don't.

I can't see how it is possible to deal with all errors inside a single global error handler. That error handler then has to know about the inner logic of every other part of the codebase, and it will also lose all context about the situtation in which the error occured. E.g. you might want to make use of a local variable in your error message, or take an entirely different action depending on which url path caused the exception, but not all of this information can be included in the exception.

Maybe map_json could be a function from 'a -> b' option * string, with the string being the post data?

Re: a separate package for JSON, that sounds fine to me, though would presumably involve a whole bunch of work with build servers and git repos and the like.

haf:
I agree that handling the branched logic is a hard topic; perhaps best addressed in #153 where we can discuss how to do control flow in an easier way... `WebPart = HttpContext -> Choice<HttpContext option, HttpContext>` where the second part ends the pipeline, perhaps? /cc @kolektiv

Albacore with automatically create a nuget for any new project you add and automatically publish it to the build server if you just place the project next to all the others.

PR #193, opened 2015-01-15
Minor documentation changes

jamesjrg:


haf:
@jamesjrg thanks, I think that wording looks good

Issue #198, opened 2015-01-30
POST without parsing body data will cause 500


xkrt:
Hi, I found a problem with Suave 0.21.1: if I ignore post data, then next request (not POST with ignored body, request after it) will return 500 (Server got itself in trouble).
Probably this issue related with #186 
This code reproduce a problem on my machine (Windows 7 x64, .net 4.5):

```fsharp
module Program

open System
open System.Net

module WebServer =
    open System.Threading
    open Suave
    open Suave.Http
    open Suave.Http.Applicatives
    open Suave.Http.Successful
    open Suave.Types
    open Suave.Logging
    open Suave.Web

    let private config = { default_config with bindings = [ HttpBinding.mk HTTP IPAddress.Any 31337us ]
                                               logger = Loggers.sane_defaults_for LogLevel.Debug }

    let private app =
        choose [ POST >>= url "/post"
                      //>>= ParsingAndControl.parse_post_data // !!! if uncomment this line all will be good
                      >>= OK "hello, post!"
                 GET >>= url "/get"
                     >>= OK "hello, get!"
                 fun ctx -> failwithf "No handler for %O" ctx.request.url ]
        >>= log config.logger log_format

    let startWith () =
        let cts = new CancellationTokenSource()
        let listening, server = web_server_async config app
        Async.Start(server, cts.Token)
        listening |> Async.RunSynchronously |> ignore
        { new System.IDisposable with member __.Dispose() = do cts.Cancel() }


[<EntryPoint>]
let main argv =
    use __ = WebServer.startWith ()

    use wc = new Net.WebClient()
    let url path = sprintf "http://%s:31337%s" System.Environment.MachineName path

    try
        wc.UploadString(url "/post", "x") |> printfn "response body: %s"
        wc.DownloadData(url "/get") |> Text.Encoding.UTF8.GetString |> printfn "response body: %s" // this will fail with 500
    with ex ->
        printfn "error: %A" ex
    0
```

ademar:
you can not ignore the post data.

xkrt:
@ademar it is bad API design, IMHO. if I forget combine my handler with parse_post_data I will have very strange behavior at runtime.

ademar:
I see how it is unexpected but it is designed that way to avoid a DOS attack where the server is busy parsing POST requests for which does not exist a handler.

Everyday in my day to day job I see IIS applications beeing DOSed to death by simply bombarding them with POST requests.

[...]

Issue #186, opened 2015-01-07
[usability] parse post data by default


haf:
```
>>= ParsingAndControl.parse_post_data
```

Should probably be opt-out, not opt-in.

Opt-in could be a value of the config that's more 'hardened' and doesn't as easily break if someone tries to put down the server.

PR #200, opened 2015-01-30
[BREAKING] (again) parse post data by default

ademar:
ref #186, #198


ademar:
what about like this ? cc: @haf @xkrt

Issue #204, opened 2015-02-09
Underscores and ticks in API and possible API review


dsyme:
Hi Suavers,

I'm considering featuring Suave in Chapter 2 or 3 of the revised Expert F# 4.0, using the getting started script here: https://github.com/SuaveIO/suave/issues/202, since it's a great example of power scripting.  WebSharper will still be featured later.

One thing that concerns me is the prevalence of underscores in the Suave API and docs.  Normally, F# APIs just don't use "_" in names.  I'm wondering, what's the rationale for this?

Going further, I wonder if you'd be opening to doing an API review with me at some point - or I could do it offline and submit a proposed patch implementing API revisions I would recommend.  Of course I would need to use the library a fair bit.

Let me know what you think, I really like the direction Suave is going and want to help make it something we'd reference in the book

Thanks
Don

ademar:
Hi Don,

I think there was not much of a rationale just that we happened to settle on that style; perhaps because `camelCase` is usually associated with OOP languages vs `snake_case` being more prevalent in languages that *look* like F# ie Python, OCaml, Ruby

I do not feel strongly either way. If @haf agrees we can revisit the naming conventions.

We are definitely open to an API review and welcome any feedback you may have, we feel honoured you have taken an interest in the project :)

Regards,

Ademar

haf:
Hi Don,

That's great news =).

I agree with Ademar in general. For me underscores is an escape from the terror of C# and also a way to make the code more amenable to non-CLR programmers, with experience in e.g. OCaml/ML or Ruby or Haskell.

It seems that you are trying to keep together the F# eco-system, and for that good cause I'd be willing to reconsider previous naming choices.

dsyme:
Fantastic, I'm enjoying looking over the library.

You are right that I guess I'd like to keep the F# ecosystem "together".  This may be quite challenging in this case as I do understand why you want to culturally move away from C# naming conventions, to put some cultural distance between you and the .NET world for this library.  Tricky.

BTW I also noticed the use of tick-marks ' in the API, e.g. stateful'.   They should definitely be removed. I adjusted the title of this issue to reflect that.

I'll send an initial "low hanging fruit" PR and make some notes for discussion

haf:
Alright, sounds good. As long as we make it a branch of its own, or a large PR, so that downstream consumers don't have to update their API usages multiple times.

dsyme:
[...]

PR #206, opened 2015-02-12
Change casing convention (API review #1)

dsyme:
As discussed in #204 there is a proposal from myself in conjunction with @haf  and @ademar to change the naming convention used in Suave to be closer to F# standards.

This PR implements a change to camelCase for at least the majority of functions in Suave.  Some are not done in the PR deliberately, e.g. 

* FOO_BAR (it's best to keep these with _) 
* foo_  identifiers that represent Lens-style functional-properties-as-values

This is a non-trivial issue since it's considerably invasive on the API. Numerous Obsolete's have been added (but it's possible Obsolete for some less common functions may have been missed)

Some other API design changes are made as well 

* foo_ become static members. Given the pattern being implemented they are intrinsically part of the type.  In their use it seems reasonable to qualify them by type name since field names are not guaranteed to be unique in any case

* Getter functions like SuaveConfig.bindings (as opposed to just using the "bindings"  property) are removed. While it's OK in principle to have these, I don't recommend them in the library design - it looks better to normalize client code to use the properties for such fundamental properties of the types. Nearly all off them were unused in the Suave code itself.

* HttpCode was moved out of "Suave.Types.Codes" and into "Suave.Types" and the associated functions made into members.  That seems the right design

* Nearly all ' functions have been renamed to clarify what the ' means. For example dir' becomes dirHomeDirectory.

I'd imagine this will generate a considerable amount of discussion and review questions. I'll add notes in the PR itself for further discussion points, please feel free to shoot away with lots of questions.

The docs have not yet been updated.


examples/Example/CounterDemo.fs
```diff
@@ -1,51 +1,51 @@
 ﻿module CounterDemo
-
```
dsyme: I'm not sure why GitHub is marking these changes, perhaps a line-ending issue

src/Suave/Logging/TraceHeader.fs
```diff
@@ -9,45 +9,38 @@ open Suave.Utils.RandomExtensions
...
     /// In ZipKin/Dapper-speak, this is the span id
-    req_id        : uint64
+    reqId        : uint64
+
     /// possibly a parent
     /// In ZipKin/Dapper-speak, this is the span parent id
-    req_parent_id : uint64 option }
+    reqParentId : uint64 option }
+
+  static member traceId_     = Property (fun x -> x.traceId) (fun v x -> { x with traceId = v })
```
kolektiv: Hope nobody minds me dropping a comment in here! This is quite an interesting one for me, as I've struggled to decide on a consistent approach to structuring lens properties/functions. At first glance static members seemed ideal, as the lens itself from Type -> Property of Type is intrinsically bound to that type. 

However, when you start to define composite lenses, where should they live? They're no longer bound to a single type. How about lenses which aren't bound to a type in the first place? (For example, lenses for the fst and snd items of a 2-tuple). You also can't extend a type alias, so when you wish to define lenses on a specialisation of a generic type, that's not possible in the same way.

It's a tricky one, and I've ended up with a mix throughout the code I write using lenses. I'm still not really delighted by that as a general rule, but if anything i'm leaning back towards lenses always being effectively a naked function (or function pair, in this formulation). Slightly off topic perhaps, but as they're being commented on here, seemed a chance to see if others had strong opinions...

src/Suave.Tests/Perf.fs
```diff
@@ -35,10 +35,10 @@ let perf_tests =
   let inline iterate times (testF : unit -> unit) =
     for i = 1 to times do testF ()
 
-  let server_factory = SuavePerfHarness("default config", run_with default_config)
+  let server_factory = SuavePerfHarness("default config", runWith defaultConfig)
 
   let getFormValue name =
-    request (fun x -> OK (HttpRequest.form x ^^ name |> Option.get))
+    request (fun x -> OK (x.queryParam name |> Option.get))
```
ademar: I think here it should be
```fsharp
let getFormValue name =
    request (fun x -> OK (x.formDataItem name |> Option.get))
```

examples/Example/CounterDemo.fs
```diff
@@ -1,51 +1,51 @@
...
+let counterDemo (req : HttpRequest) (out : Connection) =
+
+  let write i =
+    socket {
+      let msg = { id = i; data = string i; ``type`` = None }
+      do! msg |> send out
+      return! async {
+        do! Async.Sleep 100
+        return Choice1Of2 () } }
+  let q = req.queryParams
```
haf: Should we just call it `query`? Does `params` add anything?

examples/Example/Program.fs
```diff
@@ -83,41 +84,40 @@ let app =
           | None ->
             store.set "counter" 1
             >>= OK "First time")
-    basic_auth // from here on it will require authentication
-    GET >>= url "/events" >>= request (fun r -> EventSource.hand_shake (CounterDemo.counter_demo r))
-    GET >>= browse' //serves file if exists
-    GET >>= dir' //show directory listing
+    basicAuth // from here on it will require authentication
+    GET >>= url "/events" >>= request (fun r -> EventSource.handShake (CounterDemo.counterDemo r))
+    GET >>= browseHomeDirectory //serves file if exists
```
haf: Doesn't `browse` already imply that it's a directory? What about `browseHome`?

examples/Example/Program.fs
```diff
@@ -83,41 +84,40 @@ let app =
...
-      server_key       = Utils.Crypto.generate_key HttpRuntime.ServerKeyLength
-      error_handler    = default_error_handler
-      listen_timeout   = TimeSpan.FromMilliseconds 2000.
-      ct               = Async.DefaultCancellationToken
-      buffer_size      = 2048
-      max_ops          = 100
-      mime_types_map   = mime_types
-      home_folder      = None
-      compressed_files_folder = None
+      serverKey       = Utils.Crypto.generateKey HttpRuntime.ServerKeyLength
```
haf: Could we re-align this whitespace for readability?

examples/Load/Program.fs
```diff
@@ -9,16 +9,16 @@ open Suave.Http.Applicatives
 open Suave.Http.Files
 open Suave.Logging
 
-let logger = Loggers.sane_defaults_for LogLevel.Verbose
+let logger = Loggers.saneDefaultsFor LogLevel.Verbose
```
haf: When we're changing things, are we happy with this name? I find it a bit 'shoot from the hip'...

examples/Load/Program.fs
```diff
@@ -9,16 +9,16 @@ open Suave.Http.Applicatives
...
-let config : SuaveConfig = 
-  { default_config with 
+let config = 
+  { defaultConfig with 
       bindings = [ HttpBinding.mk' HTTP "127.0.0.1" 8082 ]
-      ; buffer_size = 2048
-      ; max_ops = 10000
-      ; logger = logger }
+      bufferSize = 2048
+      maxOps = 10000
```
haf: What about using unsigned integers, unless they can go below zero?

src/Experimental/Html.fs
```diff
@@ -24,75 +24,73 @@ let empty = Xml[]
...
-let html  = html' Array.empty<Attribute>
+/// Flattens an XML list
+let flatten xs = xs |> List.map (fun (Xml y) -> y) |> List.concat |> Xml
 
-let head' = tag "head"
-let head  = head' Array.empty<Attribute>
 
-let title' attr s = tag "title" attr (Xml([Text s,Xml []]))
-let title  = title' Array.empty<Attribute>
+let htmlAttr attr s = tag "html" (Array.ofList attr) s
```
haf: Wouldn't it be nicer to have `Array.toList` inside the `tag` function?

src/Experimental/Html.fs
```diff
@@ -24,75 +24,73 @@ let empty = Xml[]
...
-let (<%) f xxs = f (flatten xxs)
+let inputAttr attr = tag "input" (Array.ofList attr) empty
+let input = inputAttr [ ]
 
 /// Example
 
-let sample_page = 
-  html <% [ 
-    head <% [
+let samplePage =
```
haf: Ping @ademar - changes to syntax, removal of <% special syntax.

src/Experimental/Html.fs
```diff
@@ -116,20 +114,21 @@ let begin_element_to_string = function
        |> String.Concat
      sprintf "<%s %s>" e ats
 
-let end_element_to_string = function
+let internal endElementToString = function
```
haf: We could probably internalise a lot more of the APIs.

src/Experimental/Template.fs
```diff
@@ -40,8 +40,8 @@ type Binder = Xml -> Xml
 /// Extract an XML/HTML template from the input 'xml', in 'ns1:keyhere' form
 /// for tags that can be parsed. Replace those tags with their values as defined
 /// in the 'bindings'.
-let extract_template (nsx : string) (bindings : Map<string,Binder>) (xml : Xml) =
-  let rec extract_template' (Xml(nodelist)) (Xml(result)) : Xml =
+let extractTemplate (nsx : string) (bindings : Map<string,Binder>) (xml : Xml) =
+  let rec extractTemplateAux (Xml(nodelist)) (Xml(result)) : Xml =
```
haf: Isn't `'` a good char to use for nested private functions? Not all non-english speakers understand `auxiliary`

src/Experimental/Template.fs
```diff
@@ -67,42 +67,42 @@ let extract_template (nsx : string) (bindings : Map<string,Binder>) (xml : Xml)
...
   else stream.Write("<{0} {1}>",s,tt)
 
 /// Write the end of tag 's' to the string writer
-let write_close_tag (s : string) n a (stream : StringWriter) =
+let writeCloseTag (s : string) n a (stream : StringWriter) =
   stream.Write("</{0}>", s)
 
 /// Recursively convert the Xml structure to text into the string writer
-let rec xml_to_string1 (Xml(nodes)) (stream : StringWriter) =
+let rec xmlToString1 (Xml(nodes)) (stream : StringWriter) =
```
haf: What's with the `1` at the end? Could we change this to something more semantic?

src/Suave.OpenSSL/OpenSSL.fs
```diff
@@ -120,11 +120,11 @@ open Suave.Sockets.Connection
 let read_to_bio (con : Connection) read_bio ssl = socket {
```
haf: No change?

src/Suave.Testing/Testing.fs
```diff
@@ -9,12 +9,12 @@ Example:
 
   open Fuchu
 
-  let run_with' = run_with default_config
+  let runWith' = run_with default_config
```
haf: No change?

src/Suave.Tests/Cookie.fs
```diff
@@ -45,14 +45,14 @@ let tests =
 
     testCase "set cookie (same name) twice keeps last" <| fun _ ->
       let force = Async.RunSynchronously >> Option.get
-      let c1 = HttpCookie.mk' "a" "aa"
-      let c2 = HttpCookie.mk' "a" "bb"
+      let c1 = HttpCookie.mkSimple "a" "aa"
```
haf: `mkKV`?

src/Suave.Tests/Parsing.fs
```diff
@@ -21,21 +21,21 @@ open Suave.Testing
...
-  let post_data3 = read_text "request-2.txt"
+  let post_data1 = readBytes "request.txt"
+  let post_data2 = readText "request-1.txt"
+  let post_data3 = readText "request-2.txt"
 
-  let test_url_encoded_form field_name =
+  let test_url_encoded_form fieldName =
     request (fun r ->
-      match (HttpRequest.form r) ^^ field_name  with
+      match r.queryParam fieldName  with
```
haf: Change from `form` to query? Seems like a logic error.

src/Suave.Tests/Perf.fs
haf: Similarly to `queryParam` comment, should we just do `x.formData`? Less typing...?

src/Suave.Tests/Web.fs
```diff
@@ -29,7 +29,7 @@ let parsing_tests =
 
       // the server generates a new one with the client's as the parent?
       // is the semantics that client sends its SpanId and that
-      let expected = TraceHeader.mk (Some 7654321UL) (Some 1234567UL)
-      Assert.Equal("should parse trace id", expected.trace_id, (parse_trace_headers headers).trace_id)
-      Assert.Equal("should parse span id to parent span id", expected.req_parent_id, (parse_trace_headers headers).req_parent_id)
+      let expected = TraceHeader.create (Some 7654321UL) (Some 1234567UL)
```
haf: Should we then change `mk` throughout? Or change this back?

src/Suave/Auth.fs
```diff
@@ -34,42 +34,43 @@ let StateStoreType = "Suave.Auth"
...
-
-    cookie_state
-      { server_key      = ctx.runtime.server_key
-        cookie_name     = SessionAuthCookie
-        user_state_key  = StateStoreType
-        relative_expiry = relative_expiry
+  context (fun ctx ->
+    Log.log ctx.runtime.logger "Suave.Auth.authenticate" LogLevel.Debug "authenticating"
+
+    cookieState
```
haf: Whitespace

src/Suave/Compression.fs
```diff
@@ -75,51 +75,63 @@ module Compression =
...
-          if Globals.compressed_files_map.ContainsKey key then
-            let last_modified = get_last key
-            let cmpr_info = new FileInfo(Globals.compressed_files_map.[key])
-            if last_modified > cmpr_info.CreationTime then
-              let! new_path =  compress_file n stream compression_folder
-              Globals.compressed_files_map.[key] <- new_path
+          do! asyncWriteLn connection (String.Concat [| "Content-Encoding: "; n.ToString() |])
+          if Globals.compressedFilesMap.ContainsKey key then
+            let lastModified = getLast key
+            let cmpr_info = new FileInfo(Globals.compressedFilesMap.[key])
```
haf: Is the `Map` suffix superfluous?

src/Suave/Cookie.fs
```diff
@@ -20,15 +20,15 @@ module Cookie =
...
     | DecryptionError of Crypto.SecretboxDecryptionError
 
-  let parse_cookies (s : string) : HttpCookie list =
+  let parseCookies (s : string) : HttpCookie list =
     s.Split(';')
     |> Array.toList
     |> List.map (fun (cookie : string) ->
         let parts = cookie.Split('=')
-        HttpCookie.mk' (String.trim parts.[0]) (String.trim parts.[1]))
+        HttpCookie.mkSimple (String.trim parts.[0]) (String.trim parts.[1]))
```
haf: Create? Mk? Simple?

src/Suave/Cookie.fs
```diff
@@ -42,211 +42,224 @@ module Cookie =
...
     |> snd
 
-  module HttpRequest =
+  type HttpRequest with
 
-    let cookies (x : HttpRequest) =
+    member x.cookies =
       x.headers
       |> List.filter (fun (name, _) -> name.Equals "cookie")
-      |> List.flat_map (snd >> parse_cookies)
```
haf: Has the extra 'extension' method been removed then, too?

src/Suave/Cookie.fs
```diff
@@ -42,211 +42,224 @@ module Cookie =
...
-    unset_cookie http_cookie_name >>= unset_cookie (String.Concat [ http_cookie_name; "-client" ])
+  let unsetPair httpCookieName : WebPart =
+    unsetCookie httpCookieName >>= unsetCookie (String.Concat [ httpCookieName; "-client" ])
 
   type CookiesState =
-    { server_key      : ServerKey
-      cookie_name     : string
-      user_state_key  : string
-      relative_expiry : CookieLife
+    { serverKey      : ServerKey
```
haf: Whitespace

src/Suave/Cookie.fs
```diff
@@ -42,211 +42,224 @@ module Cookie =
...
 
   [<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
   module CookiesState =
 
-    let mk server_key cookie_name user_state_key relative_expiry secure =
-      { server_key      = server_key
-        cookie_name     = cookie_name
-        user_state_key  = user_state_key
-        relative_expiry = relative_expiry
+    let mk serverKey cookieName userStateKey relativeExpiry secure =
```
haf: mk? create?

src/Suave/Cookie.fs
```diff
@@ -42,211 +42,224 @@ module Cookie =
...
+      let plainText =
+        match readCookies csctx.serverKey csctx.cookieName ctx.response.cookies with
+        | Choice1Of2 (_, plainText) ->
+          Log.log logger "Suave.Cookie.updateCookies" LogLevel.Debug "updateCookies - existing"
+          f_plainText (Some plainText)
         | Choice2Of2 _ ->
-          Log.log logger "Suave.Cookie.update_cookies" LogLevel.Debug "update_cookies - first time"
-          f_plain_text None
+          Log.log logger "Suave.Cookie.updateCookies" LogLevel.Debug "updateCookies - first time"
+          f_plainText None
```
haf: How come you've left the `f_` prefix?

src/Suave/Cookie.fs
```diff
@@ -42,211 +42,224 @@ module Cookie =
...
+        let httpCookie, clientCookie =
+          generateCookies csctx.serverKey csctx.cookieName
+                           csctx.relativeExpiry csctx.secure
+                           plainText
+        setPair httpCookie clientCookie >>=
+          Writers.setUserData csctx.userStateKey plainText
 
-      match read_cookies csctx.server_key csctx.cookie_name (ctx.request |> HttpRequest.cookies) with
-      | Choice1Of2 (http_cookie, plain_text) ->
+      match readCookies csctx.serverKey csctx.cookieName ctx.request.cookies with
```
haf: Shouldn't we have some extra/other facility to make it explicit when we're incurring a parsing cost?

src/Suave/Http.fs
```diff
@@ -381,12 +387,12 @@ module Http =
...
         process_id //TODO: obtain connection owner via Ident protocol
-        (match Map.tryFind "user_name" ctx.user_state with Some x -> x :?> string | None -> "-")
+        (match Map.tryFind "user_name" ctx.userState with Some x -> x :?> string | None -> "-")
         (DateTime.UtcNow.ToString("dd/MMM/yyyy:hh:mm:ss %K", ci))
         (string ctx.request.``method``)
         ctx.request.url.AbsolutePath
-        ctx.request.http_version
-        (Codes.http_code ctx.response.status)
+        ctx.request.httpVersion
+        ctx.response.status.Code
```
haf: A capital property all of the sudden?

haf:
Finished reading to https://github.com/SuaveIO/suave/pull/206/files#diff-7279703bc342b7c1e54142c8ae5050b2R504 so far.

examples/Load/Program.fs
dsyme: I reckon it could do with a change :)

src/Experimental/Html.fs
dsyme: yes, done, thanks

src/Experimental/Html.fs
dsyme: Yes, it would be good to do that. I started down that path but reverted some of it because I wasn't sure. Let's do that in a follow up PR

src/Experimental/Template.fs
dsyme: I tend to avoid more or less any use of ' in F# coding these days.  I tend to use the Aux suffix, which I think is OK, but there's no particular need either way.  I've changed this one back to extractTemplate' because it's not important for the PR

src/Experimental/Template.fs
dsyme: Changed to writeXmlToStringWriter since that's what it is, could be a private writeXml too

haf: Isn't it obvious from the types that the function takes that it's to a string writer? Anyway, I'd welcome @ademar 's opinion here, with regards to making this method private, so that it can have a 'simple' name -- writing the XML surely is something the user wouldn't do?

src/Suave.OpenSSL/OpenSSL.fs
dsyme: Now fixed

src/Suave.Testing/Testing.fs
dsyme: Not sure what to call it. runWithDefaultConfig?

dsyme: Update with "runWithConfig"

haf: I mean the `'` char

src/Suave.Tests/Cookie.fs
dsyme: I've changed it to that for now.

haf:
@ademar Could you make *AppVeyor* do `gem install bundler` before `bundle exec rake`, please?

src/Suave/Auth.fs
dsyme: done,thanks

src/Suave/Compression.fs
dsyme: It's just an internal table so doesn't matter too much. Perhaps it should be changed from a ConcurrentDictionary to System.Runtime.Caching.ObjectCache or the like.

src/Suave/Cookie.fs
dsyme: Now mkKV for now as per above, but I still don't think it's ideal.

src/Suave/Cookie.fs
dsyme: done

src/Suave/Cookie.fs
dsyme: Let's talk about how we progress the "mk" pattern in a separate PR.

src/Suave/Cookie.fs
dsyme: I actually don't mind this use of "_" to name different function parameters in internal implementation code. Really it should be "fPlainText".  I think it's the fact that the "f" gets harder to see and it's really quite important.

src/Suave/Cookie.fs
dsyme: Yes, that's a recurring problem with ".cookies", ".paramData" and ".formData", all of which parse.  The options are
 
1. move to use an object instead of a record to allow these to be cached internally in the object
1. add a cache field to the record to allow these to be cached internally in the record
1. use a property name like cookieData 
1. use a method like getCookieData

I think 1 or 2 is easiest for the user - no thinking is required. 1 is really the "right" way to do things but then you lose record syntax (see https://fslang.uservoice.com/forums/245727-f-language/suggestions/6420709-allow-record-like-obj-with-newvals-syntax-for).  So 2 may be the right option.

Alternatively revert to the function HttpRequest.cookies, but it's somehow a bit clumsy in the common case.

src/Suave/Http.fs
dsyme: Changed, for consistency

src/Suave/Logging/TraceHeader.fs
dsyme: I don't think any programming language has a really great place to but items ABC composed from primitives A, B and C.  Can you send a link to the design patterns you use?  Perhaps also add comments on this thread detailing the design pattern you use: https://fslang.uservoice.com/forums/245727-f-language/suggestions/6906132-implement-first-class-lensing-lenses-in-f

examples/Example/CounterDemo.fs
dsyme: I think it's related to the "cost of parsing" question further below

src/Suave/Logging/TraceHeader.fs
kolektiv: I use (and wrote) Aether (https://github.com/xyncro/aether) for lensing in F#. The API for that is slightly non-F# ish and I'm actually considering a pending breaking change to fix that, but how it works isn't too likely to change now. It's based on the fairly naive implementation of lenses as composable get,set pairs, which is about the only option in F# short of language level changes (which, to be honest, would either have to be very feature-specific, which would make me hesitate, or very high level and general and "change to the type system" which doesn't seem too likely at this point). The naive implementation does work well enough though, and makes working with complex data structures more tractable/feasible.

Here's a (slightly nasty, but internal!) example from Hekate: https://github.com/xyncro/hekate/blob/master/src/Hekate/Hekate.fs#L177-L183 - note that the lenses are defined away from the types anyway, as they apply to type aliases. Here's a slightly nicer example where lenses are used to provide simple access to a more complex structure under the guise of some functions: https://github.com/kolektiv/freya/blob/develop/src/Freya.Machine/Configuration.fs - in that example one of the lenses being composed is defined as a static member on the type (as that's possible), but it doesn't feel like a convention that *quite* fits when other lenses won't be able to be. Also, from the consumer perspective, whether a lens is composed or not is opaque - it's just a lens, so distinguishing them, when all can be composed, etc. seems slightly awkward.

(As an aside, I'm about 95% sold on changing the functions such as getL, modL, setPL, etc. to things like Lens.get, Lens.map, and Lens.setPartial respectively. Increased verbosity and typing, but it is rather more the "F# way" - I think! I've gone more in that direction with more recent libs...)

kolektiv: Oh. as a slightly random add-on to that - the Haskell convention for lens naming (which obviously has rather different standards generally, but...) is usually the name of the property of the data-type, and a tick. So rather than my current usual standard of something like ``configLens`` it's more likely to just be ``config'``

ademar:
@haf *AppVeyor* now hangs precisly on that line.

ademar:
AppVeyor build is passing.

examples/Example/Program.fs
dsyme: We'd have to change all of these:

    val browseHomeDirectory : WebPart
    val browseFileInHomeDirectory : fileName:string -> WebPart
    val dirHomeDirectory : WebPart

to

    val browseHome : WebPart
    val browseFileInHome : fileName:string -> WebPart
    val dirHome : WebPart

or

    val browseHome : WebPart
    val browseFileHome : fileName:string -> WebPart
    val dirHome : WebPart

or

    val browseHomeDir : WebPart
    val browseFileHomeDir : fileName:string -> WebPart
    val dirHomeDir : WebPart


I think the second option is best, I've applied it, let me know what you'd like.

examples/Load/Program.fs
dsyme: (Let's do this name change in a follow-up PR)

examples/Load/Program.fs
dsyme: I think signed is ok, it's pretty normal, but either way let's do it in a follow up PR/discussion?

src/Suave.Tests/Web.fs
dsyme: I'll change it back because any change should be done consistently.

src/Suave/Cookie.fs
dsyme: "HttpRequest.cookies req" has been deprecated in favour of the property (see ongoing discussion about parsing costs though)

src/Suave/Logging/TraceHeader.fs
dsyme: Cool @kolektiv - thanks for that info.  Could you add to this to https://fslang.uservoice.com/forums/245727-f-language/suggestions/6906132-implement-first-class-lensing-lenses-in-f? Thanks

ademar:
@dsyme can we merge this ?

src/Suave/Types.fs
```diff
@@ -191,20 +208,28 @@ type HttpCookie =
     path      : string option
     domain    : string option
     secure    : bool
-    http_only : bool }
+    httpOnly  : bool }
+
+  static member nameP = (fun x -> x.name),    fun v x -> { x with name = v }
```
haf: Mix between `P` and `_` for the changed lenses.

src/Suave/Types.fs
```diff
@@ -293,62 +277,25 @@ type MimeType =
...
-module MimeType =
-
-  let mk name compression =
-    { name        = name
-      compression = compression }
-
-  let name_ =
-    (fun x -> x.name),
-    fun v (x : MimeType) -> { x with name = v }
+  static member name_ = Property (fun x -> x.name) (fun v (x : MimeType) -> { x with name = v })
```
haf: Here for example, see https://github.com/SuaveIO/suave/pull/206/files#r24666038

src/Suave/Types.fs
```diff
@@ -293,62 +277,25 @@ type MimeType =
...
-  let field_name_ =
-    (fun x -> x.field_name),
-    fun v x -> { x with field_name = v }
+  { fieldName     : string
+    fileName      : string
+    mimeType      : string
+    tempFilePath : string }
 
-  let file_name x = x.file_name
+  static member fieldName_ = Property<HttpUpload,_> (fun x -> x.fieldName) (fun v x -> { x with fieldName = v })
```
haf: Here

src/Suave/Types.fs
```diff
@@ -356,174 +303,136 @@ type Host =
...
+    rawQuery         : string
     files            : HttpUpload list
-    multipart_fields : (string * string) list
+    multiPartFields  : (string * string) list
     trace            : TraceHeader
-    is_secure        : bool
+    isSecure         : bool
     ipaddr           : IPAddress }
 
+  static member httpVersion_     = Property<HttpRequest,_> (fun x -> x.httpVersion) (fun v (x : HttpRequest) -> { x with httpVersion = v })
```
haf: Here

src/Suave/Types.fs
```diff
@@ -356,174 +303,136 @@ type Host =
...
+  member x.raw_query = x.rawQuery
+
 [<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
 module HttpRequest =
 
   open Suave.Utils
 
   let empty =
-    { http_version     = "1.1"
+    { httpVersion     = "1.1"
```
haf: Whitespace

src/Suave/Utils/Bytes.fs
```diff
@@ -1,14 +1,17 @@
...
-  open System.Text
-  open Suave.Async
+open System
+open System.IO
+open System.Text
+open Suave.Utils.Async
+
+
+
+module (* internal *) Bytes =
```
haf: +1 on making it internal

src/Suave/Utils/Crypto.fs
```diff
@@ -112,44 +112,44 @@ module Crypto =
 
       let bw  = new BinaryWriter(cipher_text)
       bw.Write iv
-      bw.Write (mk_cipher_text msg key iv)
+      bw.Write (mkCipherText msg key iv)
       bw.Flush ()
 
-      let hmac = hmac' key (cipher_text.ToArray())
+      let hmac = hmacOfBytes key (cipher_text.ToArray())
```
haf: One more rename here

src/Suave/Web.fs
```diff
@@ -68,11 +76,11 @@ module ParsingAndControl =
 
   /// Iterates over a BufferSegment list looking for a marker, data before the marker 
   /// is sent to the function select and the corresponding buffers are released
-  let scan_marker marker select connection = socket {
+  let scanMarker marker select connection = socket {
 
     match kmp_z marker connection.segments with
```
haf: One more rename here

haf:
Looking good!

src/Suave/Utils/Bytes.fs
dsyme: I ran into trouble with BufferSegment being exposed in some public API (which can probably also be made internal).  Let's use a separate PR for this whole "internal" issue.

dsyme:
I think this is almost ready to go.

The biggest question is about the naming and "parsing costs" for these:

    request.queryParams
    request.formData
    request.headers
    request.cookies

    request.queryParam key
    request.formDataItem key
    request.header key
    request.cookie key

See discussion [here](https://github.com/SuaveIO/suave/pull/206/files#r24599811) 

First, the naming is a bit inconsistent. 

Second, it seems that all of these should have their parsing costs amortized (the results of parsing computed on-demand and stored inside the request or response object), regardless of which methods/properties are used to access the information.

ademar:
Yes we should give a simpler more similar names to these two:

```
request.queryParams
request.formData
```
Maybe just `query` and `form` 

We could also discuss packing `multipart_fields` into `formData` since people usually expect them to be there

> Second, it seems that all of these should have their parsing costs amortized ..

Shall we tackle this in another PR ? I think this topic is not trivial, one could argue that if you need to parse two times is probably because you are doing it wrong ( ref: the parse post data by default argument ). Note that the modified HttpContext (the one possibly containing the parsed data) does not *trickle down* along the next route.

dsyme:
> Maybe just query and form

I don't think "query" works - it seems to convey the wrong meaning, especially in the context of F# web programming where the word "query" is so related to LINQ, query { ... } and jQuery.  

However I've updated to use it for now, so we have

    request.query
    request.form
    request.headers
    request.cookies

    request.queryParam key
    request.formData key
    request.header key
    request.cookie key

I can see the problem with amortization.

dsyme:
OK, so I think this is ready

PR #217, opened 2015-03-11
Update content.md

dsyme:
Fix underscores etc.


haf:
Great!

I'm trying to make this work btw: https://github.com/haf/sample-hello-fsharp

Perhaps you have an idea why the CLR/mono can't download (off the top of your head?) https://github.com/deis/deis/issues/3262

PR #219, opened 2015-03-20
add heroku instructions

dsyme:
This adds some instructions for using Suave + F# + Heroku.

* Please check them after merging.  

* The two "dsyme" repositories should I think be moved into https://github.com/SuaveIO to make all information related to Suave self-contained.  If you just clone them across and adjust the docs that would be great.

* The Mono 3.0 being used is the one originally used by https://github.com/aktowns/mono3-buildpack.  This was built by that person from an earlier version of Mono 3 with prefix ``app/vendor/mono3``.  We really need to be getting a ZIP-installable builds of later, up-to-date, build of Mono+F# from somewhere else - ideally mono-project.com would include these.

The instructions to use other hosting providers will be similar I suppose; most of them seem to follow a buildpack model.

### DIFF
--- a/_includes/content.md
+++ b/_includes/content.md
@@ -594,3 +594,51 @@ let webConfig =
       logger   = SuaveAdapter(logary.GetLogger "suave")
   }
 {% endhighlight %}
+
+Deploying Suave to Heroku
+----------------------------
+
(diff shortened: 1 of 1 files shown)


ademar:
This works beautifully :) Thanks!

PR #221, opened 2015-03-20
More updates

dsyme:
@haf  @ademar - the formatting is a bit messed up, can you take a look.  Here's an attempt to improve it a little


ademar:
looks somewhat better, perhaps there are too many numbered bullets

PR #222, opened 2015-03-20
more updates

dsyme:


dsyme:
@haf  - I'm calling it quits here. I see you've cloned the examples now - please merge across my latest updates and fix everything up, then I'll delete my two repos :)

ademar:
OK, we will fix it.

PR #224, opened 2015-03-21
remove old screenshot

dsyme:
This screen shot is out of date, I think it can just be removed. To be honest it looks a bit complex anyway so early in the docs for the library


haf:
@dsyme thank you, I agree

PR #226, opened 2015-03-26
[cookies] change default path to the root

theimowski:
from http://www.nczonline.net/blog/2009/05/05/http-cookies-explained/:

> The default value for the path option is the path of the URL that sent the Set-Cookie header.

So in case it's `None` the browser will save it with the path of URL


haf:
Just asking; isn't that the intended behaviour?

theimowski:
I'd like to use `Auth.authenticated` in path `/account/logon` but then the browser interprets the cookie with `path=/account/login` instead of `/`

haf:
I see =) Thanks.

PR #227, opened 2015-03-31
Feature/asp net

ademar:


src/Suave.Xsp/Library.fs
```diff
@@ -0,0 +1,178 @@
...
+
+open Suave
+open Suave.Types
+open Suave.Types.Codes
+open Suave.Http
+open Suave.Utils
+
+open Suave.Sockets
+
+module Xsp =
```
dsyme: Could you add a short summary comment on each module and type definition, and preferably each member too?  For example what Xsp stands for  (I know it's going to be eXtreme Server Pages :) ) thanks

src/Suave.Xsp/Library.fs
```diff
@@ -0,0 +1,178 @@
...
+
+  type Connection(transport : ITransport) =
+    inherit MarshalByRefObject()
```
dsyme: Could you add a comment about why this needs to be MarshalByRef?  I don't see that used so much anymore so am curious.

src/Suave.Xsp/Script.fsx
```diff
@@ -0,0 +1,8 @@
+﻿// Learn more about F# at http://fsharp.net. See the 'F# Tutorial' project
+// for more guidance on F# programming.
+
```
dsyme: Should this file be removed?

src/Suave/Http.fs
```diff
@@ -220,7 +220,7 @@ module Http =
      
 
     let not_modified : WebPart =
-      fun ctx -> { ctx with response = {status = HTTP_304; headers = []; content = Bytes [||] }} |> succeed
+      fun ctx -> { ctx with response = {status = HTTP_304; headers = []; content = Bytes [||]; writePreamble = true }} |> succeed
```
dsyme: What's this change? thanks

src/Suave.Xsp/README.txt
```diff
@@ -0,0 +1,9 @@
+
```
dsyme: This README doesn't say much, some more info here would be great, thanks

dsyme:
OK, so this lets you host ASP.NET applications using Suave? Is the main scenario on Mono/Linux?

I do wonder if this should go in a new package and project e.g. https://github.com/SuaveIO/Suave.Xsp.  Or under some kind of "contrib" directory.

ademar:
> OK, so this lets you host ASP.NET applications using Suave? Is the main scenario on Mono/Linux?

Exactly.

src/Suave/Http.fs
ademar: Sometimes we do not want Suave to send the HTTP preamble. Like in the Xsp module, the module is going to write directly to the socket and we do not want the default Suave workflow to execute. It is a little bit of a hack.

I think the need also arised when @haf was working on Server-Sent events.

ademar:
I am shooting for a Linux web hosting platform somewhat ala Heroku. 

That is why I started by implementing the Proxy module, I wanted to combine Application Request Routing (ARR) and LXC containers.

haf:
Nice, soon there will be proxies written with Suave :dancer:

src/Suave/Http.fs
haf: Can we encode it as a field of SocketTask? Are there any other than SocketTask that need to avoid writing preambles?

railcar88:
Now I get this feature.

src/Suave/Http.fs
ademar: Hum .. yeah I think you are correct .. let me explore that.

ademar: Actually `SocketTask` is not good, `startWebServer` takes a `WebPart`

PR #231, opened 2015-04-08
Use Property to mark lenses of HttpContext.

kseo:
Move HttpContext lenses from HttpContext module to HttpContext type. Change
them to static members of HttpContext type and mark them with Property to make
it explicit that they are lenses.

This patch makes HttpContext consistent with other types such as HttpRequest
and HttpResult.


haf:
@kseo Thank you!

PR #243, opened 2015-04-28
Correct Http.fsi link.

tmpreston:


haf:
@tmpreston thank you for your contribution and help!

PR #244, opened 2015-04-28
Fix Suave + Paket = ♥ example

PatrickMcDonald:
Indent the 2 lines after the `if not (File.Exists "paket.exe") then`

The example does not work without the indentation when pasted into a new script


ademar:
Thanks!

PR #245, opened 2015-05-09
Invalidate Razor file cache when file changes

tpetricek:
When using Suave in-process (F# Interactive or with automatic reloader), I wanted to be able to change a Razor template and see the change immediately on the page. This change makes this possible by tracking when the template was last changed and invalidating it if the file changes. The only tricky thing is that we need to use the time stamp as part of the Razor "template cache key" (otherwise `Razor.Parse` throws).

This changes the type of public function `loadTemplateCached`, but I suspect this should probably be marked as private, because it looks more like internals than like something that should be publicly exposed...

### DIFF
--- a/src/Suave.Razor/Library.fs
+++ b/src/Suave.Razor/Library.fs
@@ -12,29 +12,30 @@ open RazorEngine
 
 module Razor =
 
-  let private asyncMemoize f =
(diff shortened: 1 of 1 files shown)


ademar:
Thanks!

Issue #228, opened 2015-04-06
tcp request processing failed


haf:
```
[V] 2015-04-06T13:51:12.5074830Z: tcp request processing failed [Suave.Tcp.tcp_ip_server.job] exn:
System.Exception: did not find header, because header_params received None
  at Suave.Utils.Parsing.headerParams (Microsoft.FSharp.Core.FSharpOption`1 header) [0x00000] in <filename unknown>:0
  at Suave.Web+ParsingAndControl+loop@207-31.Invoke (System.Tuple`2 _arg2) [0x00000] in <filename unknown>:0
  at <StartupCode$Suave>.$SocketMonad+Bind@15-1[System.Tuple`2[Microsoft.FSharp.Collections.FSharpList`1[System.Tuple`2[System.String,System.String]],Suave.Sockets.Connection],Suave.Types+HttpContext].Invoke (Microsoft.FSharp.Core.FSharpChoice`2 _arg1) [0x00000] in <filename unknown>:0
  at Microsoft.FSharp.Control.AsyncBuilderImpl+args@787-1[Microsoft.FSharp.Core.FSharpChoice`2[System.Tuple`2[Microsoft.FSharp.Collections.FSharpList`1[System.Tuple`2[System.String,System.String]],Suave.Sockets.Connection],Suave.Sockets.SocketUtils+Error],Microsoft.FSharp.Core.FSharpChoice`2[Suave.Types+HttpContext,Suave.Sockets.SocketUtils+Error]].Invoke (Microsoft.FSharp.Core.FSharpChoice`2 a) [0x00000] in <filename unknown>:0
```

ademar:
Do you have a repo ? I suspect there is a missing (required) header in the request.

haf:
Repro: https://github.com/haf/Http.fs/commit/f8859ccf003bba5508a10424d328d346e23f146b

haf:
Run ./build.fsx

ademar:
Finding difficulty to build/run this on Windows and Linux ...

haf:
Hmm, ok... What errors are you getting?

ademar:
Never mind got it to build. The FAKE reference in the build script is
incorrect.



On Monday, April 6, 2015, Henrik Feldt <notifications@github.com> wrote:

> Hmm, ok... What errors are you getting?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/SuaveIO/suave/issues/228#issuecomment-90082302>.
>


[...]

PR #250, opened 2015-05-18
Refactor towards BAD REQUEST on parsing errors

haf:
Related to #228


haf:
We need to capture the error case as well;

```
[V] 2015-05-18T15:35:29.0210400Z: stopping tcp server, reason: 'tcpIpServer async cancelled' [Tcp.stop_tcp]
[V] 2015-05-18T15:35:29.0212120Z: stopped tcp server [Tcp.stop_tcp]
```

 To be done. But it's better than hanging forever.

ademar:
I like these changes :+1

ademar:
Although I think I would prefer a function name there rather than the operator.

haf:
@ademar I'm not sure 4bc7f3f is the right solution; because it's still not returning BAD_REQUEST properly but causing a server shut down... Insights? I'm writing a unit test that exposes this new behaviour, and then we can go around throughout the server and remove all calls to `failwith` and `failwithf`.

I'm also thinking this code could use a good fuzzer to ensure we never block/crash.

ademar:
Cool, I'll check out the branch and investigate.

haf:
@ademar I pushed a failing test

haf:
It's failing here:

```
        let! fieldName =
          (headerParams.TryLookup "name" |> Choice.map (String.trimc '"'))
          @|! "Key 'name' was not present in 'content-disposition'"
```

Because the post looks like:

```
// 8<
--/tkQKiFqMgZt:mHzua_JFrUFWHgNid
Content-Type: multipart/mixed; boundary=-idczlATz:FmvuIs'aHQSrGltky:Td
Content-Disposition: form-data; name="files"

---idczlATz:FmvuIs'aHQSrGltky:Td
Content-Disposition: file; filename="file1.txt"
Content-Type: text/plain
// >8
```

I'm posting two files under the same form field.

haf:
#251 - continuation of this

PR #252, opened 2015-05-19
Refactor toward BAD_REQUEST variant 1

ademar:


ademar:
@haf is this what you are shooting for ? https://travis-ci.org/SuaveIO/suave/builds/63204016#L213

haf:
Yes, exactly that! =)

Except of course that particular request that is failing is not a bad request, but a buggy parser for the case when a multipart-mixed is sending a two files under a single form field.

haf:
Could you solve the failing test and create a new failing test; e.g. when not sending the Host header we should also get 404 Bad Request back.

And also, I think the bad request body should contain the error message; but it probably does, but I'm not reading enough bytes from the stream :)

ademar:
Ok cool, I'll take care of it.

On Tuesday, May 19, 2015, Henrik Feldt <notifications@github.com> wrote:

> Could you solve the failing test and create a new failing test; e.g. when
> not sending the Host header we should also get 404 Bad Request back.
>
> And also, I think the bad request body should contain the error message;
> but it probably does, but I'm not reading enough bytes from the stream :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/SuaveIO/suave/pull/252#issuecomment-103603408>.
>


-- 
Ademar Gonzalez
+1-647-891-3606
http://ademar.name
https://github.com/ademar

haf:
Nice! I think it became much nicer with these changes. I'm going to test this a little bit now, and if that all works, I suggest merging and releasing v0.28. Ok with that?

haf:
Wrote release notes https://github.com/SuaveIO/suave/releases/tag/v0.28.0

haf:
Ready to tag, since I haven't gotten around to setting up Paket yet, and because @nuget team still (2 years later, or what is it?) hasn't fixed the pack bug in NuGet, I'm not pushing the pkg yet -- if Ademar gets to it before me, he can publish the release, otherwise I'm changing to paket in the repo and pushing.

ademar:
I'm uploading v0.28.0 nugets now.

haf:
Awesome! It seems I got the attention of someone on the nuget team, too, so hopefully they'll fix their client =)

I've already tweeted.

PR #259, opened 2015-05-22
Adding indexed property in http request to get form values like that of in ASP.NET web pages

tamizhvendan:
Hi,

As the title describes, this pull request adds a indexed property in the http request which enable us to get the form values from the request in a natural way as we used to do in [ASP.NET web pages](http://www.w3schools.com/aspnet/webpages_forms.asp).

image

### DIFF
--- a/src/Suave.Tests/Types.fs
+++ b/src/Suave.Tests/Types.fs
@@ -7,6 +7,8 @@ open Fuchu
 
 open Suave.Sockets
 open Suave.Types
+open System.Net.Http
+open Suave.Testing
 
 [<Tests>]
 let socket_binding =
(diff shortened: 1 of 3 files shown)


ademar:
I like this. Ideally the indexer should search all the collections `queryParam`,`formData` and `multiPartFields ` in that order. Isn't it the way ASP.NET does it ? What do you think?

tamizhvendan:
@ademar Yeah, Just now I checked the [documentation](https://msdn.microsoft.com/en-us/library/system.web.httprequestbase.item(v=vs.110).aspx). Will add the support and update this pull request.

ademar:
Thanks!

PR #264, opened 2015-06-13
prevent paket bootstrapper from requiring old .net frameworks

adamchester:
On windows 10 (10130) with only VS 2015 installed (no .net framework/runtime 2.0/3.5), the old `paket.bootstrapper.exe` requires one to install old .net framework versions.

This updated bootstrapper does not require it.


haf:
@adamchester thank you for your contribution. I merged this in the referenced commit =).

Issue #273, opened 2015-07-02
Possible issue with Random


xkrt:
Hi,

https://github.com/SuaveIO/suave/blob/master/src/Suave/Globals.fs#L5

As far I understand `Globals.random` may be used from different threads simultaneously, but System.Random class is not thread safe, see http://blogs.msdn.com/b/pfxteam/archive/2014/10/20/9434171.aspx.

Possible solution:
```fsharp
module ThreadSafeRandom =
    open System
    open System.Threading

    let private seed = ref Environment.TickCount
    let private rnd = new ThreadLocal<Random>(fun () -> Random (Interlocked.Increment seed))

    let next minInclusive maxExclusive =
        rnd.Value.Next(minInclusive, maxExclusive)

    let nextBytes buf =
        rnd.Value.NextBytes(buf)
```

https://github.com/SuaveIO/suave/blob/master/src/Suave/Utils/Crypto.fs#L50
Same possible issue with `Crypto.cryptRandom`, MSDN says it is not thread safe.

haf:
That's true; how about making a PR? Good catch!

For reference, random is used for:

 - generate a request id (multi-threaded)
 - generate the server-key on start (mostly single-threaded)

ademar:
Is it possible to write a thread-safe random generator without locks ?

xkrt:
@ademar as far I understand my solution above without locks. Will send PR a little bit later.

ademar:
AFAIK `Interlocked` still uses locks.

ademar:
Anyhow, please proceed. I  was just wondering if it was mathematically possible to code such a thing without locks.

xkrt:
@ademar well, I cant imagine how it can be anyhow possible without any "atomic read-increment-write" since every thread should get its own unque seed for `Random`.

Anyway `Interlocked.Increment` called only single time on every thread on static `Random` initialization and [MSDN](https://msdn.microsoft.com/en-us/library/sbhbke0y(v=vs.110).aspx) says that `On modern processors, the methods of the Interlocked class can often be implemented by a single instruction` so there should be no any performance problems.

ademar:
@xkrt I see what you mean. Yeah, this is a nice solution. Thanks!

PR #274, opened 2015-07-02
Globals.random replaced with ThreadSafeRandom

xkrt:
Fix for #273 

Crypto.cryptRandom not needed to fix since `RandomNumberGenerator.Create` actually return `RNGCryptoServiceProvider` which is thread safe.


ademar:
:+1:

PR #276, opened 2015-07-05
more correct handling of cancellationtoken

mastoj:
Minor update to the topshelfdemo to handle the `CancellationTokenSource` more correct. Everything works as before, but this is probably a more correct way of doing it.


ademar:
Thanks!

PR #286, opened 2015-08-13
Bug: doesn't show the first value

haf:
Output:

``` text
id: 2
data: Hello World
```

For code:

```
    GET >>= path "/events2" >>= request (fun _ -> EventSource.handShake (fun out ->
      socket {
        let msg = { id = "1"; data = "First Message"; ``type`` = None }
        do! msg |> send out
        let msg = { id = "2"; data = "Second Message"; ``type`` = None }
        do! msg |> send out
      }))
```

Ready to merge.

### DIFF
--- a/examples/Example/Program.fs
+++ b/examples/Example/Program.fs
@@ -4,9 +4,12 @@ open System
 open System.Net
 
 open Suave
+open Suave.Sockets
+open Suave.Sockets.Control
 open Suave.Logging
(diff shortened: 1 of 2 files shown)


ademar:
It works fine here

```
open: events2
message: First Message
message: Second Message
```

haf:
Yes, I've fixed the bug now.

PR #287, opened 2015-08-15
[Cookie] Ignore invalid cookies in requests

paavohuhtala:
Currently when trying to parse a malformed cookie (with a trailing comma or a one sided equals sign), `Cookie.parseCookies` will throw an `IndexOutOfRangeException`. I've added additional checks to said method to prevent this + two unit tests. Invalid cookies are silently ignored.

@ademar suggested on [Gitter](https://gitter.im/SuaveIO/suave?at=55ce71369b45e15c4264466c) that the server should reply with a `BAD_REQUEST` if parsing a cookie fails, or at least log a warning. I like both of these ideas, however they're currently impossible to implement (as far as I know) without changing the function signature and breaking backwards compatibility, since cookies are only parsed when the client asks for them and `parseCookies` has no (and IMO shouldn't have) access to the `HttpContext`.

Maybe a better solution would be to return a `Choice<HttpCookie list, unit> `or a custom discriminated union instead of a list to indicate a parse failure. The user of the API could then decide how to act. Another, stricter approach would be to parse cookies in every incoming request and automatically reject those which cannot be parsed. That could however have performance implications.

This is a temporary solution until such (if any) changes are implemented.

### DIFF
--- a/src/Suave.Tests/Cookie.fs
+++ b/src/Suave.Tests/Cookie.fs
@@ -11,7 +11,7 @@ open FsCheck
 open Tests.TestUtilities
 
 [<Tests>]
-let tests =
+let parseResultCookie =
   testList "parse result cookie" [
     testCase "parse path" <| fun _ ->
       let sample = @"st=oFqpYxbMObHvpEW!QLzedHwSZ1gZnotBs$; Path=/; HttpOnly"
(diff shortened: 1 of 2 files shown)


ademar:
Nice, thanks! If @haf agrees we can merge this.

haf:
+1 merge :). Kudos to @paavohuhtala !

PR #288, opened 2015-08-23
[WIP] Structured logging and log messages

### LINKED ISSUE #270: Consider embracing structured log messages
I thought it's time to start an issue here for discussion. I believe the issue of structured log messages can be separated from structured logging in general.

Here is an example of what structured log messages would enable:
image

Original proposal: https://twitter.com/adamchester/status/608031081351118849
Related discussion: adamchester/MessageTemplates#1
Similar discussion: https://github.com/akkadotnet/akka.net/issues/1045

adamchester:
This is a WIP for discussion. This may eventually close #270.

### DIFF
--- a/src/Suave/Logging/LogLine.fs
+++ b/src/Suave/Logging/LogLine.fs
@@ -2,6 +2,8 @@
 
 open Suave
 
+type LogTemplateFormatter = string -> (string -> obj) -> string
+
 /// When logging, write a log line like this with the source of your
 /// log line as well as a message and an optional exception.
 type LogLine =
@@ -13,22 +15,44 @@ type LogLine =
     level         : LogLevel
     /// the source of the log line, e.g. 'ModuleName.FunctionName'
     path          : string
-    /// the message that the application wants to log
-    message       : string
+    /// the template message that the application wants to log
+    template      : string
+    /// a function which can format the log template message. When the
+    /// message is pre-formatted, this is None. When the log message 
+    /// contains structured data, this function can be used to render
+    /// the message template with the data.
+    formatter     : LogTemplateFormatter option
+    /// any key-value based data to log
+    data          : Map<string, obj>
     /// an optional exception
     ``exception`` : exn option
     /// timestamp when this log line was created
     tsUTCTicks    : int64 }
-    
+
+    /// Formats the log line as a string
+    member x.formatMessage () =
+        match x.formatter with
+        | None -> x.template
+        | Some fmter -> fmter x.template x.GetDataByName
+
+    /// Gets a data value
+    member x.GetDataByName name = Option.toObj (x.data.TryFind name)
+
 [<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
 module LogLine =
   let mk path level trace ex message =
-    { message       = message
+    { template      = message
+      formatter     = None
+      data          = Map.empty
       level         = level
       path          = path
       ``exception`` = ex
       trace         = trace
       tsUTCTicks  = Globals.utcNow().Ticks }
+
+  /// assists with type inference, avoiding ugly 'box' calls for each
+  /// value item when constructing the log data map 
+  let mapLogData (data: (string*obj) list) = Map.ofList data
 
   let trace_ =
     (fun x -> x.trace),
@@ -43,5 +67,6 @@ module LogLine =
     fun v (x : LogLine) -> { x with path = v }
 
   let message_ =
-    (fun x -> x.message),
-    fun v (x : LogLine) -> { x with message = v }
+    (fun (x : LogLine) -> x.formatMessage()),
+    fun v (x : LogLine) -> { x with template = v
+                                    formatter = None }
--- a/src/Suave/Http.fs
+++ b/src/Suave/Http.fs
@@ -401,10 +401,12 @@ module Http =
         ctx.response.status.code
         "0"
 
-    let log (logger : Logger) (formatter : HttpContext -> string) (ctx : HttpContext) =
+    let log (logger : Logger) (contextFormatter : HttpContext -> string) (ctx : HttpContext) =
       logger.Log LogLevel.Debug <| fun _ ->
         { trace         = ctx.request.trace
-          message       = formatter ctx
+          template      = String.Empty
+          formatter     = Some (fun _ _ -> contextFormatter(ctx))
+          data          = Map.empty
           level         = LogLevel.Debug
           path          = "Suave.Http.web-requests"
           ``exception`` = None
--- a/src/Suave/Logging/Logger.fs
+++ b/src/Suave/Logging/Logger.fs
@@ -21,7 +21,7 @@ module Loggers =
     // [I] 2014-04-05T12:34:56Z: Hello World! [my.sample.app]
     "[" + Char.ToUpperInvariant(line.level.ToString().[0]).ToString() + "] " +
     (DateTime(line.tsUTCTicks, DateTimeKind.Utc).ToString("o")) + ": " +
-    line.message + " [" + line.path + "]" +
+    line.formatMessage() + " [" + line.path + "]" +
     (match line.``exception`` with | Some e -> " exn:\n" + e.ToString() | None -> "")
 
   /// Log a line with the given format, printing the current time in UTC ISO-8601 format
--- a/src/Suave/Tcp.fs
+++ b/src/Suave/Tcp.fs
@@ -175,9 +175,15 @@ let startTcpIpServerAsync (bufferSize  : int, maxConcurrentOps : int)
       acceptingConnections.Complete startData |> ignore
 
       logger.Log LogLevel.Info <| fun _ ->
+        let cancelReqMsg = if token.IsCancellationRequested then ", cancellation requested" else ""
         { path          = "Suave.Tcp.tcpIpServer"
           trace         = TraceHeader.empty
-          message       = sprintf "listener started in %O%s" startData (if token.IsCancellationRequested then ", cancellation requested" else "")
+          template      = "listener started in {@startData}{cancelReqMsg:l}"
+          data          = Map.ofList [  "startData",     box startData
+                                        "cancelReqMsg",  box cancelReqMsg ]
+          formatter     = Some (fun t getValue ->
+                                    // TODO: bring in template->string formatter
+                                    sprintf "listener started in %O%O" (getValue "startData") (getValue "cancelReqMsg"))
           level         = LogLevel.Info
           ``exception`` = None
           tsUTCTicks    = Globals.utcNow().Ticks }


src/Suave/Logging/LogLine.fs
```diff
@@ -13,23 +15,45 @@ type LogLine =
...
+      data          = Map.empty
       level         = level
       path          = path
       ``exception`` = ex
       trace         = trace
       tsUTCTicks  = Globals.utcNow().Ticks }
 
+  /// assists with type inference, avoiding ugly 'box' calls for each
+  /// value item when constructing the log data map 
+  let mapLogData (data: (string*obj) list) = Map.ofList data
```
adamchester: This doesn't seem to help.

haf: The next version is likely to have something like this:

https://github.com/logary/logary/blob/feature/object-model/src/Logary/DataModel.fs#L8-L19

Is the reason you don't like it, that it boxes the types?

adamchester: Yes, I don't particularly like the ugly `box` syntax, and looking for an alternative. Not sure how much it really matters, but if structured log messages are used all over the place, a short succinct syntax seems important.

[For example](https://github.com/SuaveIO/suave/pull/288/files#diff-c035cf27914d9f92ba7a592fd960fd94R182):

```fsharp
data = Map.ofList [  "startData", box startData; "cancelReqMsg",  box cancelReqMsg ]  
```

adamchester: Also, regarding the structure values, [MessageTemplates](https://github.com/adamchester/MessageTemplates/blob/05b3bec99b3953e56ee04e5ad485d08cded85389/FsMessageTemplates/MessageTemplates.fsi#L66-L76) and Serilog capture the structures like this.

haf: I think getting a nice API is a good idea. It's easy to provide overloads that transform primitive and common F# CLR values into the object model we both linked to, but not as easy when they are complex without letting the author of the type extend the type with an extra serialisation method.

```
module Value =
  let ofCLRValue (data : ^a) =
   // ...
// and 
let addData (m : Message) (key : String) (data : ^a) =
 Lens.setLens (Message.DataLens key) (Value.ofCLRValue data)
```

Could that work? I'm thinking we can provide the (bind, apply) applicatives for a new Value monad, like Chrion does, which lets you compose the values properly; thereby both defining the wire-format as well as letting MessageTemplates use all of the structure.

As for the structure; MessageTemplates is using a subset of the Value type that I linked to. How about we use the superset of mine in this library? That will allow supporting metrics easier, like is Logary's prerogative.

adamchester: OK, the problem with this is
  1. I don't have a lot of experience in it, and;
  1. I'm not sure how much benefit it provides yet

I'm willing to experiment with this as it sounds interesting, but I'm going to need your help.

Thoughts?

haf: Sure, I'm here to help. First though, what was your initial plan to get around the boxing issue?

adamchester: Well, the annoying part was being required to manually specify the `box`, which ensures type inference does not interfere with the list/array of structured log values.

In static class members, we have the ability to use `[<ParamsArray>]`, which makes it even easier. Since the types are going to be serialised or stringified into log structures, I'm OK with them being implicitly boxed (but your idea above is interesting too).

haf: But how would you go from a boxed object of arbitrary type to a serialised representation in the object model linked to above? (either yours or mine) You'd have to do a `match x with | :? bool as b -> ScalarValue x | :? string as s -> ...` etc, probably with [no little amount of reflection](https://github.com/logary/logary/blob/master/src/Logary/Formatting.fs#L27-L69) for the generic types of List, Map, Queue etc?

Let me give you some code...

haf: I just read through your code, and it's fairly heavy on the reflection side. Could you give me an overview of it? Is it:

 1. Parse template t
 1. Capture
   1. For each binding in t
   1. destructure p maching t's position or name
 1. For each binding in t, format the object model's destructured value as a string into the stream

As for the destructuring, I think we might be able to improve some on the heuristics with my suggestion above.

screen shot 2015-08-27 at 09 19 10

adamchester: Firstly, I am [not proud of that method](https://twitter.com/adamchester/status/632693196615610368), however I'd also like to point out, it looks a lot nicer if you view it like [this](https://twitter.com/hmemcpy/status/632805273380372480/photo/1):
cmgtctrxaaapxad png-large

Yes we can definitely do better (e.g. remove need for reflection) if we make the log message creator figure it all out up front, however I would have to weigh the cost/benefit to know for sure :)

I am not (yet, at least) proposing bringing anything from 'MessageTemplates' to Suave. Having said that, I *am* quite a fan of the Serilog approach to structured log messages, but I'm trying to keep an open mind right now, and keep myself open to alternative approaches for Suave.

Serilog and MessageTemplates goes to great lengths to ensure the captured structures are immediately copied to immutable structures, so that once rendering takes place, there is no risk of anything having been changed. Critically, it does this in a way that puts *minimal* burden on the log message writer/creator. See [here](https://github.com/serilog/serilog/wiki/Structured-Data).

IMO, it should remain dead simple to construct the log message template for a generic logging library. However, in the context of Suave, I'd love to explore other possible approaches too, which may push some of the burden back onto the Suave log message creator, in order to reduce complexity of capturing the logged messages. That burden may be so small, that I end up totally convinced it's not worth a message template library; actually this is why I haven't really proposed it yet.

haf: Awesome gif :)

I jotted down some notes, based on the Chiron library: https://github.com/adamchester/MessageTemplates/pull/16

adamchester: :+1: This looks amazing, give me some days to understand this :construction:

haf: I've added the implementation to the object-model branch. The idea I have for suave is that Suave will take a github-dependency https://fsprojects.github.io/Paket/github-dependencies.html on the ObjectModel.fs file and then do a replace on the namespace. That way the object model is sufficiently expressive for all implementing projects, and Logary can provide a reflected/cached adapter implementation with a single line of code for any project that uses the file.

adamchester: OK - there are key questions I still have, and I think I need your help.

MessageTemplates & Serilog takes the approach that the log message creator can provide any object that follows some basic rules for "destructuring". You would like to avoid the destructuring complexity (and associated reflection involved) and replace it with message template construction inconvenience (just how much of an inconvenience remains to be seen).

In the MessageTemplates/Serilog model, the log message creator would do something like this:
```fsharp
log.Information "listener started in {startData}ms" startData
```

In your proposed model (inc. Logary), the log message creator would do something like this:
```fsharp
log.Information("listener started in {???} ms", convertToStructure ("startData", startData))`
// additionally, some small amount of static member constraints that allow
// startData to be converted into your object model, this must be written manually.
```

In your proposed _structured log message_ model, you haven't yet solved the issue of how to store the string template messages, and how to parse and match them to the provided arguments. Is that correct, or have I missed something?

I don't yet understand how the 'lens' complexity is justified here, could you explain or point me to something that might help clear this up for me?

haf: 1. Moving complexity around? Not really. You can still have an overload that takes an anonymous object and reflects on it. This way though, you get speed in the default case, speed if you implement ToValue or slow reflection in any other case.
 2. `{???}` is equivalent to `{startData}` because `startData` is positional, just like ???
 3. In the proposed model, inc Logary, the log message creator uses the API to add Fields that contain values, but you can invoke Message.fieldWithUnit from `logger.Information` just as well, so you don't need the ugly `convertToStructure`, because that's implicit in `Message.fieldWithUnit` since it sets a Field-value in the Message structure. The cognitive dissonance you're experiencing is that you don't have anonymous objects in F#, but you're still thinking in terms of them. But yes, if you have records, you'd either need to provide a ToValue method on it OR you can let it default to reflection, using the existing infrastructure that you've written. I suggest you write it this way.
 4. The lens doesn't add complexity, imo. It adds structure and simplicity as it separates how you change Value values from how they are structured, at the point of the call-site. And this is all about complexity *at the call site*, right?

In the proposed model I left out the parsing, since I figured we'd use your work for it; your parsing was good. The storage for the template is [the message string](https://github.com/logary/logary/blob/695ca9df777503bb9a477e4136bce25dbc84878b/src/Logary/DataModel.fs#L478); compared to your work, we delay the parsing until we need it – which we do at the point when we output the value into a Target (Sink). That way you can use the hash of the template to create a histogram of logged messages, without the templated values changing the uniqueness.

Does this make sense?

adamchester: OK most of this makes sense, at first glance.

Is there any reason why the property names ("instant" and "time" is your example) need to be specified twice? They are specified first in the template message, and again as a tuple via the `<+>` (Message.fieldWithUnit). Are you thinking that positional matching (like String.Format or Printf) is more brittle?

haf: @adamchester Specified twice: not really that many reasons; if you can make it work with positional parameters that don't lose their types and thus keep their static method lookup. Also you'd need to run parsing directly with positional parameters, lest the field values become without names (but they could be named their positions, so perhaps not necessarily).

I think we're getting somewhere in this design space :)

adamchester: I think we are getting somewhere too :)

Parsing the template before it leaves the call site has benefits:
 1. Ability to match message fields to their names (positionally), without being required to specify the names twice.
 2. Ability to warn the message author earlier if there is a problem.

The downsides I can think of:
 1. overhead of parsing OR complexity of caching them. This is a fairly small overhead, and it's already been fairly well optimised with lots of ugly code and tests. Serilog, for example, actually caches the parsed template messages to avoid this overhead. You would need to be doing many thousands of log messages before it becomes an issue worth considering (which seems quite possible with Suave and Logary and certainly MessageTemplates in general). If parsed templates were cached, we have the added complexity, but we can learn from the early mistakes [Serilog made](https://github.com/serilog/serilog/issues/523).

haf:
@adamchester This work is starting to trickle down into Logary. Thought you'd like to know :) :+1:

adamchester:
Link? :)

haf:
@adamchester https://github.com/logary/logary/pull/90 here

haf:
Logging has been merged now; a facade in front of Logary. We're considering doing more to add pretty printing in a *very* basic format to that Facade, which then will be much easier to update across libraries. Perhaps EmParsec.

PR #294, opened 2015-08-29
[owin] Add sample OWIN app to the example app

panesofglass:
@haf I added a simple OWIN app that should "just work." However, it is not completing. Can you spot what's wrong?

### DIFF
--- a/src/Suave/Owin.fs
+++ b/src/Suave/Owin.fs
@@ -282,6 +282,8 @@ module OwinAppFunc =
   type internal OwinDictionary(initialState) =
     let cts = new CancellationTokenSource()
     let state : HttpContext ref = ref initialState
+    let responseHeaders = new Dictionary<string, string[]>(StringComparer.OrdinalIgnoreCase)
+    let responseStream = new IO.MemoryStream()
 
     let req l =
       HttpContext.request_ >--> l <--> untyped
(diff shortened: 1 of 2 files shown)


src/Suave/Owin.fs
```diff
@@ -333,13 +345,14 @@ module OwinAppFunc =
...
+        // TODO: provide a Stream that wraps the following:
+        //let impl conn = AsyncSocket.transferStream conn v
+        //SocketTask impl
+        // TODO: may also provide a Stream that continuously writes to the HttpContext.
         let bs = Lens.getPartialOrElse HttpContent.BytesPLens [||] x
-        ms.Write(bs, 0, bs.Length) // hmm? flag that we have a stream?
-        upcast ms),
-      (fun v x ->
-        let impl conn = AsyncSocket.transferStream conn v
-        SocketTask impl)
```
panesofglass: I assume that the above two lines trigger sending headers, as well as fires the `OnSendingHeaders` callback.

haf: Well, not at the moment; Suave first builds the HttpResult data structure; as you can see if you look in `SocketTask`, it takes a closure it can call later. So [down here](https://github.com/SuaveIO/suave/pull/294/files#diff-b41f275d6cdfe9f95ef8dd4c3cf71946R487), the Web.fs server-implementation checks that we have a SocketTask.

What's needed, however, is to make the OWIN structure into a state machine with states `initial` and `headersSent` (at the very least).

So the above MemoryStream was just to make it so that the typed interface was fulfilled. With your example app, I can start with the state machine.

PR #295, opened 2015-08-31
[owin] adding test cases

panesofglass:
Work in progress


panesofglass:
@haf, sorry I didn't get more done. I took a little time to look over the changes, then got stuck trying to figure out why everything seemed to crater whenever I set the reason phrase. I assume you have something that throws `invalidOp`, though the reported exception is about object disposal.

haf:
@panesofglass No worries.

It doesn't work because of this line:

``` fsharp
OwinConstants.responseReasonPhrase, HttpContext.response_ <--> untyped // TODO: add support for modifying phrasing to Core?
```

Because `response_` is `Lens<HttpContext, HttpResult>`, but OWIN is to `..., string>`.

src/Suave.Tests/Owin.fs
```diff
@@ -148,5 +159,190 @@ let endToEnd =
...
+      let noContent (env : OwinEnvironment) =
+        env.[OwinConstants.responseStatusCode] <- box 204
+        async.Return ()
+      
+      let setCustomHeaders = OwinMidFunc(fun next -> OwinAppFunc(fun env ->
+          let responseHeaders : IDictionary<string, string[]> = unbox env.[OwinConstants.requestHeaders]
+          responseHeaders.["X-Custom-Before"] <- [|"Before OWIN"|]
+          next.Invoke(env).ContinueWith(fun _ ->
+            let responseHeaders : IDictionary<string, string[]> = unbox env.[OwinConstants.requestHeaders]
+            responseHeaders.["X-Custom-After"] <- [|"After OWIN"|]
```
panesofglass: IIRC, this _should_ fail unless the middleware buffers the response itself. I need to double-check.

src/Suave.Tests/Owin.fs
```diff
@@ -148,5 +159,190 @@ let endToEnd =
...
+          )
+        ))
+
+      let composedApp =
+        path "/owin"
+          >>= OwinApp.ofApp noContent
+          // NOTE: really need some sort of splitter, as found in Freya.
+          >>= OwinApp.ofMidFunc setCustomHeaders
```
panesofglass: IIRC, this _should_ fail unless the middleware buffers the response itself. I need to double-check.

src/Suave.Tests/Owin.fs
```diff
@@ -148,5 +159,190 @@ let endToEnd =
...
+      let noContent (env : OwinEnvironment) =
+        env.[OwinConstants.responseStatusCode] <- box 204
+        async.Return ()
+      
+      let setCustomHeaders = OwinMidFunc(fun next -> OwinAppFunc(fun env ->
+          let responseHeaders : IDictionary<string, string[]> = unbox env.[OwinConstants.requestHeaders]
+          responseHeaders.["X-Custom-Before"] <- [|"Before OWIN"|]
+          next.Invoke(env).ContinueWith(fun _ ->
+            let responseHeaders : IDictionary<string, string[]> = unbox env.[OwinConstants.requestHeaders]
+            responseHeaders.["X-Custom-After"] <- [|"After OWIN"|]
```
panesofglass: IIRC, this _should_ fail unless the middleware buffers the response itself. I need to double-check.

src/Suave.Tests/Owin.fs
```diff
@@ -148,5 +159,190 @@ let endToEnd =
...
+
+    testCase "OWIN middleware run after OWIN app" <| fun _ ->
+      let noContent (env : OwinEnvironment) =
+        env.[OwinConstants.responseStatusCode] <- box 204
+        async.Return ()
+      
+      let setCustomHeaderAfter = OwinMidFunc(fun next -> OwinAppFunc(fun env ->
+          next.Invoke(env).ContinueWith(fun _ ->
+            let responseHeaders : IDictionary<string, string[]> = unbox env.[OwinConstants.requestHeaders]
+            responseHeaders.["X-Custom-After"] <- [|"After OWIN"|]
```
panesofglass: IIRC, this _should_ fail unless the middleware buffers the response itself. I need to double-check.

src/Suave.Tests/Owin.fs
```diff
@@ -148,5 +159,177 @@ let endToEnd =
...
+    testCase "Composed OWIN security middleware and app" <| fun _ ->
+      let noContent = OwinAppFunc(fun env ->
+        env.[OwinConstants.responseStatusCode] <- box 204
+        Threading.Tasks.Task.FromResult() :> Threading.Tasks.Task
+        )
+      
+      let basicAuthMidFunc = OwinMidFunc(fun next -> OwinAppFunc(fun env ->
+        let requestHeaders : IDictionary<string, string[]> = unbox env.[OwinConstants.requestHeaders]
+        // NOTE: currently fails b/c headers are not using StringComparer.OrdinalIgnoreCase.
+        let authHeader = requestHeaders.["Authorization"].[0].Split([|' '|])
```
panesofglass: This test highlights the importance of using `StringComparer.OrdinalIgnoreCase`, as most OWIN apps probably use the casing common in the HTTP RFCs.

src/Suave.Tests/Owin.fs
```diff
@@ -148,5 +159,177 @@ let endToEnd =
...
+        //eq "owin.RequestPathBase" requestPathBase "/owin"
+        let requestPath : string = unbox env.[OwinConstants.requestPath]
+        printfn "owin.RequestPath is %s" requestPath
+        if requestPath <> "/app" then
+          env.[OwinConstants.responseStatusCode] <- box 404
+        else () // 200 OK
+        async.Return ()
+
+      let composedApp =
+        pathRegex "/owin(/.+)*" >>= OwinApp.ofApp ok
```
panesofglass: Composing an app with a `path`, `pathRegex`, or `pathScan` should probably work the same way as Katana's `IAppBuilder.Map`.

src/Suave.Tests/Owin.fs
```diff
@@ -148,5 +159,177 @@ let endToEnd =
...
+        async.Return ()
+
+      let composedApp =
+        pathScan "/owin/%s" (fun path -> OwinApp.ofApp (fun env -> async {
+          env.[OwinConstants.requestPathBase] <- box "/owin"
+          env.[OwinConstants.requestPath] <- box ("/" + path)
+          do! ok env
+          env.[OwinConstants.requestPathBase] <- box ""
+          env.[OwinConstants.requestPath] <- box ("/owin/" + path)
+          })
```
panesofglass: The above shows what would need to occur in order to map the `path` to the `owin.RequestPathBase`. I think `path` in Suave is meant to be used differently, so perhaps this could be moved to the Owin module?

src/Suave/Owin.fs
```diff
@@ -686,7 +687,9 @@ module OwinApp =
 
       let impl (conn, response) : SocketOp<unit> = socket {
         // disposable because it buffers what the OwinApp writes to the stream
-        use wrapper = new OwinDictionary({ ctx with response = response })
+        // also, OWIN expects the default HTTP status code to be 200 OK
+        // could also set the status code after calling SocketOp.ofAsync if the value is not set
+        use wrapper = new OwinDictionary({ ctx with response = { response with status = HTTP_200 } })
```
panesofglass: Set the expected, default status code for OWIN apps.

haf: But it can succeed without returning 200 OK, right?

src/Suave.Tests/Owin.fs
haf: Ah, yes, we could do this, but with a new combinator that shifts the path. Is this something OWIN apps use regularly?

haf:
Ok, let's merge this PR into the OWIN branch.

src/Suave/Owin.fs
panesofglass: If by "succeed" you mean "complete," then yes. However, the default for OWIN is `200 OK`. So if you mount an OWIN app that does nothing but complete, the response should send back `200 OK`.

src/Suave.Tests/Owin.fs
panesofglass: I don't know if it is used regularly, but several frameworks use it in their `IAppBuilder` extensions for Katana, e.g. SignalR and WebSharper.

panesofglass: In general, `IAppBuilder.Map` uses this, and the intent is that any path manipulation that occurs outside the OWIN `AppFunc` should be pushed into the `"owin.RequestPathBase"` value.

src/Suave.Tests/Owin.fs
panesofglass: Since I didn't state it explicitly, my understanding as to why this test fails is that I used `"Authorization"` rather than `"authorization"`, as the `KeyNotFoundException` is thrown, and `"authorization"` exists in the headers dictionary.

panesofglass: Also, the OWIN spec lists use of `OrdinalIgnoreCase` as a `MUST`. I looked through the `DeltaDictionary` but couldn't figure out how to switch out key comparison. I assume `Map<_,_>` uses the case-sensitive, structural comparison.

PR #301, opened 2015-09-17
[owin] Implemented StringComparison.OrdinalIgnoreCase

panesofglass:
@haf, I have another version at https://github.com/panesofglass/suave/tree/feature/owin-ci-list that uses a `list<KeyValuePair<string, 'T>>` rather than a `Map<OwinKey, 'T>`, in case that is of interest. Both branches solve the same problem and include the same test changes.

### DIFF
--- a/src/Suave.Tests/Owin.fs
+++ b/src/Suave.Tests/Owin.fs
@@ -31,72 +31,73 @@ let throws msg matcher fn =
 [<Tests>]
 let unit =
   let create (m : (string * string) list) =
-    OwinApp.DeltaDictionary(m)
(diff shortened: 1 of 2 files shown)


haf:
@panesofglass Thanks, let's get this into the feature branch. What's the status on the tests? Do you want to continue working on the tests or are you feeling ready?

panesofglass:
I should keep adding some tests, e.g. check for all required keys, optional keys, URL reconstruction, etc. Sorry for the delays.

panesofglass:
Also, I should note that setting response reason phrases should probably warn but not fail, so long as an app can set it without breaking the server. Suave does that fine. I'm sure other servers also refuse to allow custom response reason phrases.

PR #302, opened 2015-09-18
Accept mixed-case in websockets upgrade header

tpetricek:
IE sends `Upgrade: Websocket`, and according to the RFC, this is even allowed:

    3.   An |Upgrade| header field containing the value "websocket",
        treated as an ASCII case-insensitive value.

Looks like most of Suave codebase uses `ToLower()` rather than `String.Compare(..., StringComparison.InvariantIgnoreCase)`, so I followed the same style.

Also, it would not compile without the change in `State.fs` - but that's what I got when I run `paket restore` with the current `paket.lock` file, so I'm not sure what's happening here!

### DIFF
--- a/src/Suave/WebSocket.fs
+++ b/src/Suave/WebSocket.fs
@@ -196,7 +196,7 @@ module WebSocket =
     let r = ctx.request
     if r.``method`` <> HttpMethod.GET then
       return! RequestErrors.METHOD_NOT_ALLOWED "Method not allowed" ctx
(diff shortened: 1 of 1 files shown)


haf:
Sorry about this taking some time. I'm sorting out the FsPickler breakage, then you rebase and we merge :)

haf:
@tpetricek Could you please rebase. Will merge then.

tpetricek:
Updated!

haf:
@tpetricek Thanks for your support!

PR #303, opened 2015-09-24
Update FsPickler components

eiriktsarpalis:
Updates FsPickler related code to work with releases >= 1.3.0. Some simplifications have also been made.

### DIFF
--- a/src/Suave/State.fs
+++ b/src/Suave/State.fs
@@ -21,18 +21,13 @@ module CookieStateStore =
   [<Literal>]
   let StateCookie = "st"
 
-  let private encodeMap (map : Map<string, obj>) =
-    let pickler = FsPickler.CreateBinary ()
-    use ms = new MemoryStream()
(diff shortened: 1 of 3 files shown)


src/Suave/State.fs
```diff
@@ -21,18 +21,13 @@ module CookieStateStore =
...
-    ms.ToArray()
-
-  let private decodeMap bytes : Map<string, obj> =
-    let pickler = FsPickler.CreateBinary ()
-    use ms = new MemoryStream()
-    ms.Write (bytes, 0, bytes.Length)
-    ms.Seek (0L, SeekOrigin.Begin) |> ignore
-    pickler.Deserialize ms
+
+  let private binarySerializer = FsPickler.CreateBinarySerializer ()
```
haf: Are the serialisers thread safe?

eiriktsarpalis: Yep.

Issue #297, opened 2015-09-01
authenticateBasic doesnt store userName in userState


xkrt:
Hi!

On successful basic authentication `userName` not stored in `ctx.userState`. I want to use `userName` for logging for now and maybe for some business logic in future.

`Http.Authentication` module:
```fsharp
let authenticateBasic f (ctx : HttpContext) =
  let p = ctx.request
  match p.header "authorization" with
  | Choice1Of2 header ->
    let (typ, username, password) = parseAuthenticationToken header
    if (typ.Equals("basic")) && f (username, password) then
      fail // ! makes possible to run next 'already authenticated' webparts, but how to store userName?
    else
      challenge { ctx with userState = ctx.userState |> Map.add UserNameKey (box username) }
  | Choice2Of2 _ ->
    challenge ctx
```

ademar:
We have an issue for this, see #123

Issue #123, opened 2014-07-28
authenticate_basic should write to the `userData` map & make it available to 'downstream' web parts


ademar:
There are a couple of things wrong with `authenticate_basic`. The way the function is used feels wrong.

```fsharp
let protected_area = choose [
  authenticate_basic verify_credentials;
  route_1;
  route_2;
  ..
   ]
```
I think the following usage is better.

```fsharp
let protected_area =
  authenticate_basic verify_credentials <| choose [ route_1; route_2;  ..  ]
```
`authenticate_basic` also adds an entry `"user_name"` in the `user_state` dictionary, in the current implementation that value is not accessible inside the protected routes and it should.

PR #305, opened 2015-10-06
Authentication.authenticateBasic pass username to protected webpart

xkrt:
Related to https://github.com/SuaveIO/suave/issues/297 and https://github.com/SuaveIO/suave/issues/123

### DIFF
--- a/examples/Example/Program.fs
+++ b/examples/Example/Program.fs
@@ -87,41 +87,41 @@ let app =
           | None ->
             store.set "counter" 1
             >>= OK "First time")
-    basicAuth // from here on it will require authentication
-    // surf to: http://localhost:8082/es.html to view the ES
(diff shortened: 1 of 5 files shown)


ademar:
Thanks!

xkrt:
Can you release this please?
I think related issues may be closed.

haf:
@xkrt Great. Now this is a breaking change, so do we want an overload without the protected web part?

xkrt:
@haf well, my opinion is that Suave have version 0.3 and this means that you may not pay too many attention to backward compatabilty. This also means that API will be less swell. So I think we do not want old-style overload:)

ademar:
We are changing the function signature so it will break only their builds which shouldn't be a problem.

haf:
Hmm... OK.

Issue #293, opened 2015-08-28
Headers duplication problem


xkrt:
Hi, I have following code:

```
Authentication.authenticateBasic (flip List.contains authorizedUsers)
    >>= Writers.setHeader "WWW-Authenticate" "Basic realm=\"my custom message\""
```
And got this response:
```http
HTTP/1.1 401 Unauthorized
Server: Suave/0.30.0.0 (http://suave.io)
Date: Fri, 28 Aug 2015 09:02:45 GMT
WWW-Authenticate: Basic realm="my custom message"
WWW-Authenticate: Basic realm="protected"
Content-Length: 42

No permission -- see authorization schemes
```

As you can see `WWW-Authenticate` duplicated, first - my custom message, second - Suave built in message. I just want override built in message with my custom.

I think reason for this is:
```fsharp
type HttpResult =
  { status  : HttpCode
    headers : (string * string) list
    content : HttpContent
    writePreamble : bool }
```
`headers` have type `(string * string) list`, not `Map<string,string>` (or some Map that preserve order if it important).

xkrt:
Just find http://stackoverflow.com/questions/4371328/are-duplicate-http-response-headers-acceptable
and this http://tools.ietf.org/html/rfc2617#section-4.6

As far as I understand response may have more than one `WWW-Authenticate` header, but with different authentication schemes and user may choose between them.

Ok, I test this with Chrome and it show me my custom message, not Suave built in.

But duplication of other headers may still can be a problem.

xkrt:
My current workaround for prevent `WWW-Authenticate` duplication:
```fsharp
[<AutoOpen>]
module Utils =
    /// remove all headers with specified key
    let unsetHeader key (ctx: Suave.Types.HttpContext) =
        { ctx with response = { ctx.response with headers = List.filter (fst >> (<>) key) ctx.response.headers } }
        |> succeed

Authentication.authenticateBasic (flip List.contains authorizedUsers)
	>>= unsetHeader "WWW-Authenticate"                                 // Authentication.authenticateBasic always set `Basic realm="protected"`
	>>= Writers.setHeader "WWW-Authenticate" (sprintf "Basic realm=\"%s\"" unauthorizedMessage)
	>>= Response.response HTTP_401 [||]
```

haf:
How about a `Writers.putHeader` with overwrite-existing semantics?

xkrt:
@haf it will be great

ademar:
[...]

PR #308, opened 2015-10-09
Writers.setHeader semantic changed to replace + Writers.putHeader

xkrt:
Related to https://github.com/SuaveIO/suave/issues/293

### DIFF
--- a/src/Suave/Http.fs
+++ b/src/Suave/Http.fs
@@ -109,7 +109,11 @@ module Http =
     open System
 
     let setHeader key value (ctx : HttpContext) =
-      { ctx with response = { ctx.response with headers = (key, value) :: ctx.response.headers } }
+      { ctx with response = { ctx.response with headers = (key, value) :: (ctx.response.headers |> List.filter (fun (k,_) -> k <> key))  } }
+      |> succeed
+
+    let putHeader key value (ctx : HttpContext) =
+      { ctx with response = { ctx.response with headers = List.append ctx.response.headers [key, value] } }
       |> succeed
 
     let setUserData key value (ctx : HttpContext) =
--- a/src/Suave/Http.fs
+++ b/src/Suave/Http.fs
@@ -112,7 +112,7 @@ module Http =
       { ctx with response = { ctx.response with headers = (key, value) :: (ctx.response.headers |> List.filter (fun (k,_) -> k <> key))  } }
       |> succeed
 
-    let putHeader key value (ctx : HttpContext) =
+    let addHeader key value (ctx : HttpContext) =
       { ctx with response = { ctx.response with headers = List.append ctx.response.headers [key, value] } }
       |> succeed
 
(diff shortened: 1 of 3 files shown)


xkrt:
Looks like this is problem with mono implementation of `HttpClient` (which use Suave.Testing).
Proof:

Add to `Example.Program`:
```fsharp
GET >>= path "/mytest" >>= Writers.putHeader "X-Custom-Header" "first" >>= Writers.putHeader "X-Custom-Header" "second" >>= OK "mytest"
```

Testing client code:
```fsharp
var client = new HttpClient();
var request = new HttpRequestMessage(HttpMethod.Get, new Uri("http://test-host:8082/mytest"));
var resp = client.SendAsync(request).Result;
foreach (var header in resp.Headers)
    Console.WriteLine("{0}: {1}", header.Key, string.Join(", ", header.Value));
```

On Windows with target framework 4.0 I got:
```
X-Custom-Header: first, second
Date: Fri, 09 Oct 2015 13:20:13 GMT
Server: Suave/0.0.0.0, (http://suave.io)
```

But under Ubuntu with official `mono-complete` package  4.0.4.1-0xamarin1 I got:
```
Server: Suave/0.0.0.0, (http://suave.io)
Date: Fri, 09 Oct 2015 13:07:45 GMT
X-Custom-Header: second
```

Guys, what to do with this?

ademar:
I'm not sure what to do with that :confused:

xkrt:
I reported issue to mono https://bugzilla.xamarin.com/show_bug.cgi?id=34715

haf:
@xkrt Can you please read it from the socket like we do in some other tests, instead?

xkrt:
@haf done:)

src/Suave/Http.fs
```diff
@@ -109,7 +109,11 @@ module Http =
     open System
 
     let setHeader key value (ctx : HttpContext) =
-      { ctx with response = { ctx.response with headers = (key, value) :: ctx.response.headers } }
+      { ctx with response = { ctx.response with headers = (key, value) :: (ctx.response.headers |> List.filter (fun (k,_) -> k <> key))  } }
+      |> succeed
+
+    let putHeader key value (ctx : HttpContext) =
```
haf: How about `addHeader` instead? Put denotes overwriting to me, similar to how `set` works.

src/Suave/Http.fsi
```diff
@@ -91,9 +91,12 @@ module Http =
   /// Functions have signature f :: params... -> HttpContext -> HttpContext.
   module Writers =
 
-    /// Sets a header with the key and value specified
+    /// Adds or replace a headers with the key and value specified
     val setHeader : key:string -> value:string -> WebPart
```
haf: I'm not onboard with the doc change, as it seems more accurate before. How about "Ensures that the `key` header has the given `value`", instead? And for `addHeader` (renamed from `putHeader`): "Adds the header `key` with the given `value` to the list of returned headers, even if that header already exists"

haf:
@xkrt great fix!

src/Suave/Http.fsi
xkrt: @haf I agree with both remarks

haf:
@xkrt Awesome, let's merge! =)

PR #315, opened 2015-10-25
setup appveyor x64 builds

ademar:
The purpose of this PR is to have appveyor run tests for 32 and 64 bits platforms.


examples/Asp.NET/Asp.NET.fsproj
```diff
@@ -1,91 +1,114 @@
...
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+    <DebugType>pdbonly</DebugType>
+    <Optimize>true</Optimize>
+    <Tailcalls>true</Tailcalls>
+    <OutputPath>bin\Release\</OutputPath>
+    <DefineConstants>TRACE</DefineConstants>
+    <WarningLevel>3</WarningLevel>
+    <PlatformTarget>AnyCPU</PlatformTarget>
+    <DocumentationFile>bin\Release\Asp.NET.XML</DocumentationFile>
+    <Prefer32Bit>true</Prefer32Bit>
```
haf: Should we prefer 32 bits?

haf:
Looking good, read through the changes and it seems to be mostly csproj changes.

examples/Asp.NET/Asp.NET.fsproj
ademar: No we do not. But the flag has always been there (see in master); does not seem to have any effect.

Issue #314, opened 2015-10-24
concurrency bug with Suave.Libuv 0.33 


vietnt:
with 50 concurrent connections (1-49 ok), after 2k-5k requests (random) there is a 70-80% chance (multi-thread-bug?) Suave.Libuv exited with the following exception:

    [I] 2015-10-24T14:30:50.4658733Z: listener started in 87.584 ms with binding 127
    .0.0.1:8083 [Suave.Tcp.tcpIpServer]
    [I] 2015-10-24T14:31:34.3782624Z: could not start LibUvSocket [LibUvServer.run]
    exn:
    System.NullReferenceException: Object reference not set to an instance of an obj
    ect.
       at Suave.LibUv.Native.uv_run(IntPtr loop, Int32 mode)
       at Suave.LibUv.Tcp.LibUvSocket.run(IntPtr server)
       at Suave.LibUv.Tcp.LibUvServer.run()
    [I] 2015-10-24T14:31:34.3812648Z: exiting server. [LibUvServer.run]

code:

    [<EntryPoint>]
    let main argv = 
        let cfg = { defaultConfig with tcpServerFactory = LibUvServerFactory(); maxOps = 1000 }
        startWebServer cfg (OK "hello world")
        System.Console.ReadLine() |> ignore
        0

weigHttp command

    -n 10000 -c 50 localhost:8083

see: #313 for weigHttp-download-link

haf:
Might this be because the buffer manager doesn't auto-expand and libuv code doesn't take that into account?

ademar:
Most likely something is being collected by the GC that it shouldn't. I was
trying to get away without pinning handles but apparently it is needed.

I'll go over this this afternoon.

On Saturday, 24 October 2015, Henrik Feldt <notifications@github.com> wrote:

> Might this be because the buffer manager doesn't auto-expand and libuv
> code doesn't take that into account?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/SuaveIO/suave/issues/314#issuecomment-150823889>.
>


-- 
Ademar Gonzalez
+1-647-891-3606
http://ademar.name
https://github.com/ademar

ademar:
I can not trigger it on Windows 10 (32-bits) no matter how big the numbers. 

I'm looking now for a 64bit install to test this.

vietnt:
sorry, here is the environment I use:
- Suave.Libuv 0.33 (64bit - uncheck prefer 32bit)
[...]

PR #317, opened 2015-10-29
A fix for #314

ademar:


ademar:
@haf want to review this PR? Should fix #314 cc: @vietnt @neoeinstein

neoeinstein:
I've completed a trivial test of 10,000,000 requests at 50 concurrent connections over about 2 hours with no stability issues. I'll take a closer look at the diff in the morning.

src/Suave.LibUv/Tcp.fs
```diff
@@ -141,11 +145,11 @@ type LibUvTransport(pool : ConcurrentPool<OperationPair>,loop : IntPtr,client :
 
   member this.initialize() =
     this.uv_close_cb <- uv_close_cb(this.closeCallback)
+    this.pin <- GCHandle.Alloc(this)
 
-  member this.shutdown() =
-    Async.FromContinuations <| fun (ok, _, _) ->
-      synchronizationContext.Post(SendOrPostCallback(fun o -> this.close ok),null)
-      synchronizationContext.Send()
+  member this.shutdown () = Async.FromContinuations <| fun (ok, _, _) ->
```
haf: Whitespace/formatting? Move Async down or unindent lines below? Preferably move down Async.

src/Suave.LibUv/Tcp.fs
```diff
@@ -211,17 +220,15 @@ type LibUvSocket(pool : ConcurrentPool<OperationPair>,logger, serveClient, ip, l
       uv_tcp_init(loop, client) |> checkStatus
 
       if (uv_accept(server, client) = 0) then
-        let transport = new LibUvTransport(pool,loop,client,this.synchronizationContext,logger)
+        let transport = new LibUvTransport(pool,loop,client,synchronizationContext,logger)
         transport.initialize()
         Async.Start <|
```
haf: We should avoid these  `<|` unless they make it much clearer.

src/Suave.LibUv/Tcp.fs
```diff
@@ -298,7 +308,7 @@ type LibUvServer(maxConcurrentOps, bufferManager, logger : Logger,
     Marshal.FreeCoTaskMem handle
 
   member this.closeHandler (handle : IntPtr) (arg : IntPtr) =
-    uv_close(handle,uv_close_cb(this.closeHandlerCallback))
+    uv_close(handle,this.uv_close_cb_handler)
 
   member this.uv_stop_loop (_ : IntPtr) =
```
haf: Should we name F# methods in this manner? I understand why p-invoke method should; but F# ones?

src/Suave/Sockets/Connection.fs
```diff
@@ -25,7 +25,7 @@ module Connection =
```
haf: This file was made executable in this commit.

src/Suave/Sockets/Connection.fs
```diff
@@ -25,7 +25,7 @@ module Connection =
 
   let empty : Connection =
     { socketBinding = SocketBinding.mk IPAddress.IPv6Loopback 8083us
-      transport     = { socket = null; readArgs = null; writeArgs = null }
```
haf: Would it make sense to default to the default transport?

src/Suave/Sockets/ITransport.fs
```diff
@@ -1,5 +1,7 @@
 ﻿namespace Suave.Sockets
```
haf: This file was made executable in this commit.

src/Suave/Sockets/ITransport.fs
```diff
@@ -1,5 +1,7 @@
...
 
+[<AllowNullLiteral>]
```
haf: Should we add a comment to this effect on ITransport; F# devs may not be expecting nulls, and usage of this interface goes all the way up to the user during configuration.

src/Suave/Sockets/TcpTransport.fs
```diff
@@ -3,13 +3,29 @@
 open System
 open System.Net.Sockets
 
-type TcpTransport =
-  { socket    : Socket
-    readArgs  : SocketAsyncEventArgs
-    writeArgs : SocketAsyncEventArgs
-  }
+type TcpTransport(acceptArgs: SocketAsyncEventArgs, a : ConcurrentPool<SocketAsyncEventArgs>,b : ConcurrentPool<SocketAsyncEventArgs>,c : ConcurrentPool<SocketAsyncEventArgs>) =
```
haf: Line breaks?

src/Suave/Sockets/TcpTransport.fs
```diff
@@ -3,13 +3,29 @@
...
+  let writeArgs = c.Pop()
+  let shutdownSocket _ =
+    try
+      if socket <> null then
+        try
+          socket.Shutdown(SocketShutdown.Both)
+        with _ -> ()
+        socket.Close ()
+        socket.Dispose ()
+    with _ -> ()
```
haf: To consider in the future: have a counter checking exceptions? Or we could use the System.Diagnostics global counter during a stress tests; because if this happens a lot, it can be a drag on Suave's performance.

src/Suave/Tcp.fs
```diff
@@ -118,7 +117,7 @@ let job logger
     }
   try
     use! oo = Async.OnCancel (fun () -> intern "disconnected client (async cancel)"
-                                        Async.RunSynchronously shutdownTransport)
+                                        Async.RunSynchronously <| transport.shutdown())
```
haf: `<|` same here

PR #318, opened 2015-10-30
Add `Connection: keep-alive` header when keep-alive is requested

neoeinstein:
Suave does support HTTP 1.1 keep-alive, keeping the socket connection open by default. However, it does not send back a `Connection: Keep-Alive` header in the response for any client. For backward compatibility with HTTP 1.0 clients, such as some benchmarking tools, a keep-alive header is required to keep the connection open. This adds a keep-alive header in the response when the requester has explicitly requested a keep-alive connection and the client is HTTP 1.0.

### DIFF
--- a/src/Suave.Tests/Web.fs
+++ b/src/Suave.Tests/Web.fs
@@ -34,3 +34,29 @@ let parsing_tests (_: SuaveConfig) =
       Assert.Equal("should parse trace id", expected.traceId, (parseTraceHeaders headers).traceId)
       Assert.Equal("should parse span id to parent span id", expected.reqParentId, (parseTraceHeaders headers).reqParentId)
     ]
+
+[<Tests>]
+let keepAliveTests =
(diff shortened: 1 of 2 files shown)


ademar:
Nice find. However the fix is incorrect. 

What about if the programmer wants the request to end? With this fix there is no oportunity for him to override the header and close the connection.

I think  you should add the header to the response record somewhere around here https://github.com/SuaveIO/suave/blob/master/src/Suave/Web.fs#L547 before passing the context to the WebPart; then if the programmer wants to end the request he can override it in the WebPart.

Do you follow me? :smiley:

neoeinstein:
I follow you. I'll rework the PR to take that into account.

ademar:
This is good, tests are passing now. Do you mind adding a little test for this ?

neoeinstein:
Certainly. Should be added shortly.

ademar:
This looks great. Thank you very much!

Issue #331, opened 2015-11-09
Suave live reload


haf:
[Spike](https://github.com/SuaveIO/suave/tree/feature/live-reload) – something to get you started if you want to pick this issue to work at.

Whenever the user changes a .fs file, the server should recompile and reload. It needs to be in a separate process to make it work consistently with different DLL versions.

JohnDoeKyrgyz:
I have a prototype of this working. I created an application called Suave.ActiveServer that can be run from the command line to continuously build and serve an application under development. My implementation can be found here: [https://github.com/JohnDoeKyrgyz/suave/commit/10a728251349ca2684fdc1c7a140d235fed8a945] 

This is just a prototype so far. I would be interested in some code review / feedback / suggestions to improve this before it becomes a pull request.

haf:
@JohnDoeKyrgyz Nice :) Give me a couple of evenings and I'll do a review of it!

forki:
We already have suave live reloading working with a FAKE build around it. See fsreveal

haf:
I have seen it, a lot of this is fetched from there. But it depends on FAKE and it only works with fsx files not DLLs.

Banashek:
Just for reference, since usually live-reloading is only done in development, I handle it like this: https://github.com/Banashek/Universal-FSharp-Samples/blob/master/UniversalLiveReloading/src/App.fs . And use a port-incrementing approach to avoid socketexceptions only when the `DEVELOPMENT=true` environment variable is set.

This is with dotnet core, so I can run `dotnet watch run` and it will auto restart the suave application, though on another port since the old port is not freed early enough to rebind to.

PR #333, opened 2015-11-09
[Reloaded] towards #331

haf:
From
- https://github.com/tpetricek/suave-xplat-gettingstarted/blob/master/build.fsx
- https://github.com/fsharp/FAKE/blob/master/src/app/FakeLib/ChangeWatcher.fs
- https://github.com/fsharp/FAKE/commit/68c22d13b25675b7be65391d7792e456fb2c962e#commitcomment-14286610


haf:
/cc @tpetricek – I'm porting your fsx code to something that will also support different dll versions. 
/cc @polytypic – More Hopac, haven't gotten nacks right yet. Is this in the right direction?

Feedback (on this non-compiling state) welcome :)

haf:
Closing in favour of #360

src/Suave.Reloaded/Reloaded.fs
```diff
@@ -0,0 +1,175 @@
...
+        typ      = typ
+        value    = domain.CreateInstanceAndUnwrap(typ.Assembly.FullName, typ.FullName) :?> 't }
+
+    member x.Value = x.value
+
+    interface IDisposable with
+      member x.Dispose () =
+        AppDomain.Unload x.domain
+
+  let watcher (globPattern : string) : Hopac.Ch<FileSystemEventArgs> =
```
polytypic: One typically does not expose channel. Rather, one typically exposes alts or jobs or functions that return alts or jobs.

src/Suave.Reloaded/Reloaded.fs
```diff
@@ -0,0 +1,175 @@
...
+
+        let runningState = ScriptState.Create() |> reloadScript (serverConfig homeFolder)
+        let _, server = startWebServerAsync serverConfig state.app
+        let runningServer = Hopac.Extensions.Async.toAlt server |> queue
+        return! loop (runningServer, runningState)
+      }
+      and loop (runningServer, state) : Alt<unit> =
+        let changes = watcher "*.*"
+        Alt.choose [ changes >>% reconfigure
+                     cancellation ]
```
polytypic: The basic loop here seems ok.  However, instead of cancellation or nacks, why not just send a message (in the broad sense) to request termination.  The [`IAsyncDisposable`](http://hopac.github.io/Hopac/Hopac.html#def:type%20Hopac.IAsyncDisposable) documentation might give some ideas.

Issue #311, opened 2015-10-19
Interest in choosing a different operator for >>= ? 


amcguier:
I was curious if there'd be any interest in doing a breaking version change to redefine WebPart composition to use a different operator than >>=, or alternatively changing the calling conventions so that >>= works more like a standard monadic bind? 

The current operator is inlined, which prevents the operator from being overloaded, which means if you're doing something like:
http://fsharpforfunandprofit.com/posts/recipe-part2/
you can't introduce the more standard usage of the operator into the namespace as Suave already defines it in an incompatible way.

Just for reference sake, the was some previous discussion of this in:
https://github.com/SuaveIO/suave/issues/298

I'd be glad to help out with making the change, or testing it!

haf:
Redefining is too large a change that will break too much, despite our version number. But we could define an alternative module that exposes different things than now. What do you think of that?

amcguier:
An alternative module would work fine for me. The only trick would be the >>= as defined couldn't leak into the namespace of that module or we'd be back to square one, which might be hard since it's defined in the primary interface file.

To be clear, when I suggested the change, I wasn't thinking of continuing with the >>= operator and redefining it's behavior like as described #298, I was thinking more keeping the behavior the same and choosing another operator for the same operation that didn't have another accepted meaning. A find and replace, if you will.

Either way would work for my use case really. Thanks for the consideration.

haf:
I'm thinking that you could create a module that manually aliases everything from the present module of `>>=` and then just open that instead? And go through all [<AutoOpen>] to ensure they are at least one step under a child module of the module containing `>>=`.

haf:
Do you understand what I mean? You can do a module like:

``` fsharp
module SuaveAltBindindings

let (*>>) = Suave.Http.bind
let never = Suave.Http.never
let ...
```

And then in your code, you never open `Suave.Http` which contains the original operator.

amcguier:
That does make sense. I'll take a look at whipping this up early next week maybe and see how I do.

dubiousdavid:
I would suggest using `>=>` in place of `>>=`, as that is the correct operator, while offering `>>=` in addition, for a time, to give folks time to make the switch in their code.

haf:
Closing in favor of a PR, when it comes.

PR #335, opened 2015-11-12
Features/operator redefinition

amcguier:
Hi Folk,

This is a first stab at providing a slightly cleaner infix situation, referencing item #311. This cleans the namespace and allows other libraries to use monadic bind >>= in the standard way. 

I'm pretty open to suggestion about this one. One alternative approach might be to have all the named prefix functions remain in the primary namespace, and move both my rebound infix operators, as well as the original Suave defined operators into separate modules where folk could use them (or not) without opening them into the global namespace by default. 

That approach would require a little refactoring of the core modules to add an open statement for the default operators, but it might be the cleanest way to move forward.

Thoughts?


src/Suave/HttpAltBindings.fs
```diff
@@ -0,0 +1,48 @@
...
+  let inline warbler f a = Http.warbler f a
+  let inline cnst x = Http.cnst x
+  let cond = Http.cond
+
+
+  //Here we expose Suave's WebPart composition with the more standard >=> operator
+  //Which frees other libraries to use >>= in the more tranditional way.
+
+
+  //Using a dummy type here so that our operator definition doesn't automatically smash
```
haf: Could you make this a real xmldoc comment?

src/Suave/HttpAltBindings.fs
```diff
@@ -0,0 +1,48 @@
...
+    static member inline (>=>) (BindIt,first) = fun second -> first >=> second
+    static member inline (<=<) (BindIt,first) = fun second -> first <=< second
+
+
+  let inline (>=>) first second = (BindIt, first) >=> second
+  let inline (<=<) first second = (BindIt, first) <=< second
+
+
+
+  module Response = Http.Response
```
haf: Could you copy-n-paste the docs from Http.fs?

haf:
"[...] move both my rebound infix operators, as well as the original Suave defined operators into separate modules where folk could use them (or not) without opening them into the global namespace by default."

But the operators aren't in the global namespace today, are they? They are in the Suave.Http module, and you can choose not to open it?

About the type member overloading; what would a downstream consumer override look like? You cannot attach more static members to the BindIt type, right? So instead you'd have to implement the inline operator on the type in question; but then you can't do that for framework types, like Choice, right?

haf:
We could do `open Suave.Operators`, perhaps? /cc @ademar @panesofglass @dsyme How do you think that affect the public API? Considering we'd be moving towards v1 API lockdown sometime now.

src/Suave/HttpAltBindings.fs
amcguier: Will do

panesofglass:
As I recall, we used a child `Operators` module in F#x, where applicable. I don't think we did that for Freya, but we also moved away from having so many operators as a general rule. I don't know what's the _right_ thing to do here, though. I don't personally mind having operators made available when I open a module.

ademar:
@haf I am OK moving the operators to its own name space; do not know about the name though. The idea would be to phase them out eventually and promote usage of the symbols defined in this PR.

ademar:
Perhaps use `Suave.Operators` for this module instead of `HttpAltBindings`.

haf:
So currently we have:

```
open Suave
open Suave.Http
open Suave.Types
...
```

as the default. With that change it would be

```

```
open Suave
open Suave.Http
open Suave.Types
open Suave.Operators
..
```

I think that is too much; if we perform this change, how about we make it AutoOpen in Suave and for the rest one would have to choose a submodule of the Sauve namespace? I.e.

```
open Suave // also opens Operators by default
// we want to start a web server from this place/module
open Suave.Web
```

I was also wondering if you @amcguier would consider answering my questions above?

**Comment by ademar** [comment]
We should spend some time thinking about the name spaces; I do not quite like the way they are now.

For example I feel that `Suave.Http` is a bad name.

**Comment by ademar** [comment]
Also `Suave.Types` types should probably inhabit the default name space.

**Comment by panesofglass** [comment]
Agree about `Suave.Types`. I don't know about `Suave.Http`. Do you have any plans for opening up other networking protocols? What if I want to run a custom protocol over UDP? If you want to pursue that, then `Suave.Http` makes sense. Also, where does the Web Sockets and Server Sent Events live?

**Comment by ademar** [comment]
>  Also, where does the Web Sockets and Server Sent Events live?

I haven't made my mind yet but I'm thinking of moving them one level up. So we do.
```fsharp
open Suave.EventSource
```
Instead of.
```fsharp
open Suave.Http.EventSource
```
What do you think ? I would like to have less open statements in general.

panesofglass:
@ademar, well, `EventSource` is really part of `HTTP`, so I would think it sort of stays put, but I could also see the argument that one may want to _only_ use `EventSource` and not the rest of HTTP. Really sort of depends on how it's built out, I think. If you don't need the rest of `Http`, bump it up.

amcguier:
@haf Sorry, I kind of lost them in the ongoing discussion.

1. You're right, they're not open into the global namespace. However, there are lots of commonly functions in the Http namespace that you otherwise have to access in a fully qualified manner which means adding a module alias prefix every time you want to access one if you don't want the operators. Honestly I don't have strong feelings about this one way or another.

2. I put the BindIt trick in there more as a novelty example. In this case it probably isn't as useful. That little hack around is useful if the operator is already defined into your namespace because you can add an overload without masking the existing implementation. It's most useful when someone has already defined the operator you're trying to use and you want to preserve their definition and add your own without masking the original operator. A downstream consumer could repeat the same trick and fall back through to the Suave definition, assuming they opened it into their namespace.

3. As far as not having the additional namespace I can see that. You don't want it to be too complicated. I think your idea of having them auto-open into Suave is a good compromise as long as there aren't other types that are essential to consuming Suave that are only accessible from that namespace.

Did I get them all? If not, let me know. :-)

haf:
Ref #340

haf:
Ref #342

amcguier:
@haf If the operators are going to get moved into a namespace that makes it easy not to open them into the project namespace, I can just kill this PR and use that change. I'm too stuck on the changes I made anyway.

Thoughts?

haf:
@amcguier Yes, alright. Let's go for the namespace refactor PR. You're welcome to chime in on the refactor branch.

PR #334, opened 2015-11-12
Enable build and run on coreclr for suave

KevinRansom:
Folks

Okay ... this PR contains nuget.exe ... it is used to get the correct platform dependencies.  Personally, I think it is the only way that will work until paket is updated to properly understand the nuget v3.0 workflows, but you are welcome to modify it however you see fit.  I promise I won't be offended.

Okay now how to build and how to run:

* The existing desktop build should be unaffected, and produces results in build\debug and build\release.
* To build for coreclr:
 in the root of the suave repo type:
````
tools\paket install
````
* This gets the correct dependencies for suave, it also gets the f# compiler for all 4, supported platforms and fsharp.data fot the deployment script

* In the src\suave directory type:
````
msbuild suave.fsproj /p:Configuration=Release /p:TargetPlatform-coreclr
````
* This builds suave using the coreclr version of the F# compiler.

* Once build is complete, in the root directory of the suave repo type:
````
tools\deployall
````
* This generates 4 folders below src\suave\bin\deployment, one for each supported platform.

I have only actually tested on ubuntu, since that was the purpose of this effort, to see if it would work.  I am very confident that since that works, the desktop ones will work too.  The OSX deployment is provided for the adventurous amongst you who also use Macs in coffee shops.

To run it on any of the supported platforms, copy all of the deployment folder to a subdirectory on the target platform.  I called it Suave for example.

From the compiler directory from the packages folder copy all of the files to a separate Compiler directory, don't smash them into a single directory, they use different versions of their dependencies.

The compiler is very ... very alpha quality, although it compiles F# good, the command line experience needs a lot of polishing.  Bug reports are very welcome.

We need to work out how to deal with reference assemblies and implementation assemblies and deployment and blah ... and blah ... and blah ...

I will probably work on a script and a sample of how to do it including references:

However, this really does work, there is no smoke or mirrors just vanilla Ubuntu and coreclr and F#.

Kevin

PR #336, opened 2015-11-13
Enable build and run on coreclr for suave – Paket

forki:
For now this will only be a series of tests on the CI.


haf:
Closing in favor of #334

PR #340, opened 2015-11-20
Namespace refactoring

ademar:
Most awesome refactoring evah.

Brings Http stuff closer to Suave.


haf:
I see you've created a module per monad; how about organising the operators like this, around it?

https://github.com/haf/YoLo/blob/master/YoLo.fs

ademar:
I'll do that.

On Saturday, 21 November 2015, Henrik Feldt <notifications@github.com>
wrote:

> I see you've created a module per monad; how about organising the
> operators like this, around it?
>
> https://github.com/haf/YoLo/blob/master/YoLo.fs
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/SuaveIO/suave/pull/340#issuecomment-158672907>.
>


-- 
Ademar Gonzalez
+1-647-891-3606
http://ademar.name
https://github.com/ademar

haf:
Can you find a place for

```

module Operators =

  let (|@) opt errorMsg =
    match opt with
    | Some x -> Choice1Of2 x
    | None -> Choice2Of2 errorMsg

  let (||@) c errorMsg =
    match c with
    | Choice1Of2 x -> Choice1Of2 x
    | Choice2Of2 y -> Choice2Of2 errorMsg
```

In your refactor? I'm currently in the process of moving all utility modules to either using YoLo as a paket-referenced file, or being internal, or being documented. I.e. no undocumented public, nor internal which can be better maintained with my YoLo file that I use myself all over the place.

haf:
Let's make very sure we don't remove public APIs in merging this; but just deprecate them for a while.

ademar:
@haf Henrik I do not like the unnecessary hocus-pocus with the YoLo.fs file

Can you bring it under our tree? And give it a serious name ?

haf:
I've added it under our tree. The name means that it's frustrating to write these helpers for all projects. We can rename it if you want?

haf:
Starting to look good! I would like to try this on my app that use most of Suave's surface area before merge, alright?

ademar:
@haf what do you think of renaming module `Suave.Applicatives` to `Suave.Filters` ?

haf:
Yes, I think that’s a good idea.

haf:
Releasing v0.34.0-nsrefactor so I can test/iterate on this from the consumer-side.

haf:
Need to have `[<Obsolete( .. )>]` attributes for anything moved.

ademar:
> Need to have [<Obsolete( .. )>] attributes for anything moved.

I do not follow, where exactly would you place the Obsoletes?

haf:
Mirror the API so that it still compiles for people migrating to newer versions.

ademar:
@haf last commit is an attempt to address that but I'm not very happy with it and tempted to revert it. 

In particular do not like that opening now `Suave.Http` will pollute the environment with the the old operators, forcing the programmer to write (while the deprecated operators still exist):

```fsharp
open Suave.Http            // introduce old >>=
open Suave.Operators // <-- necessary to override old >>=
```
I'm afraid it may be source of confusion ...

haf:
Thanks. It's a start, but we'll need to expose all public apis, not just the modules. And I think we should also include help about topics like the one you're discussing above, as a part of the obsoletion attributes. Finally, I think it's easier to refer to the new modules from the old, instead of duplicating the code. Like:

``` fsharp
module Http =
  [<Obsolete>]
  module Applicatives = global.Suave.Filters
```

ademar:
> but we'll need to expose all public apis, not just the modules.

OK, you are welcome to try your hand at it. I think it is not practical and a waste of time.

I agree we can alleviate the problem by writing correct and abundant documentation for the release.

haf:
Ok, I will. I think it's totally worth the time to keep users happy.

haf:
We can use the Classic/Deprecated namespace to provide the Obsolete compiler warnings. And state in release notes: "`open Suave.Deprecated` to get help from the compiler to perform the migration; remove when you're done". That way the new API is not "opt-in" by placing after all other opens.

haf:
I also think we should move the Tcp factories to the main namespace. Opening `Suave.Tcp` will e.g. give access to `createPools` globally; but you'd just want to specify the default tcp factory. What do you think?

ademar:
Yes, agree on moving `DefaultTcpServerFactory` to the main name space, same level as ` DefaulTlsProvider`.

All this complication arises from renaming `Suave.Types` -> `Suave.Http` so perhaps I should scratch that change?

haf:
No, I think it's generally a good idea. I spent a couple of hours today mapping out the changes, writing docs for the new APIs and creating fsi files. I'm also redoing video material for F#.tv due to these changes, so I want them done properly (once and for all ;))

ademar:
now we have nuget packages on appveyor, see https://ci.appveyor.com/project/AdemarGonzalez/suave/build/1.0.1003/job/ryc7lh7pwrjplfsw/artifacts

ademar:
AppVeyor NuGet feed:

https://ci.appveyor.com/nuget/suave-wwf4q3qudake

ademar:
@haf do you want to do anything else here ? I want to merge it.

We can still do something later regarding backward compatibility but I would like to have this set of changes in master.

haf:
Yes I was planning on doing some work on this today.

haf:
WebPart is generically typed now?

ademar:
@haf 

```fsharp
type WebPart = WebPart<HttpContext>
```

ademar:
@haf between one of the reasons it does not compile is that Http.fsi is missing the `AutoOpen` attribute, see:  https://github.com/SuaveIO/suave/blob/20e9d0849a8369ad01125c13af5eaeeb60d7ae1f/src/Suave/Http.fs#L4

haf:
Thanks, there are just a few declarations left. And then there's this (list of breaking changes, besides the modules):

dsc_0712

haf:
What about `Suave.WebPart` and `Suave.Operators`. I think it would be much more consistent to have `Suave.WebPart.Operators`

ademar:
I definitely prefer here Suave.Operators.

haf:
Should we add the validation operators to the same namespace then? Like `||@` and `|@` for option and choice respectively?

ademar:
Sure I think that's OK.

haf:
How about we start writing the release notes to help with the migration? I can start...

ademar:
Go ahead, I'll help you later. Been up the last 24 hours migrating a
datacenter. My legs are killing me :/

On Friday, 18 December 2015, Henrik Feldt <notifications@github.com> wrote:

> How about we start writing the release notes to help with the migration? I
> can start...
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/SuaveIO/suave/pull/340#issuecomment-165845089>.
>


-- 
Ademar Gonzalez
+1-647-891-3606
http://ademar.name
https://github.com/ademar

haf:
I need to fix OWIN properly before v1.0, if that's OK, but I think we should merge this branch.

haf:
And we should have a press release, too :)

PR #351, opened 2015-12-09
Add .editorconfig file.  See http://editorconfig.org/

jhugard:
Add editorconfig file to help compliance with style guide.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Suave Style Guide
+[*.{fs,fsi,fsx}]
+charset = utf-8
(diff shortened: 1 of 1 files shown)


haf:
+1. Any reason for all the changes to the sln file?

jhugard:
No, largely unintentional.  Only significant change is to add neuroscientist to the solution.  Sorry, thought about cleaning that up, but was in a hurry this morning.

It's probably just an artifact of using different Visual Studio patch levels or versions. Note that the differences are mostly GUIDs changing case.

jhugard:
neuroscientist == iPhone spelling correction of .editorconfig:p

jhugard:
Updated to just include the .editorconfig file (no solution file changes).

ademar:
Thanks!

PR #352, opened 2015-12-09
Fix bug: lenght -> length.  Fixes Xsp issue when Content-Length is -1.

jhugard:
Noticed a misspelling which leads to incorrect handling in Xsp when Content-Length is -1: the value is ignored due to misspelling of length as "lenght".


haf:
Looking good. Waiting for travis.

haf:
All green! Thank you and welcome as a contributor to Suave! I hope we'll see more great stuff from you!

Issue #357, opened 2015-12-11
Freya and Suave


kolektiv:
Referencing #350, i've had a play with it and tried to see what's breaking. First of all, I tried to run a super simple Freya implementation of an OwinAppFunc, which was of this:

```fsharp
let m =
    "Hello World!"B

let ok =
        Freya.Optic.set Response.reasonPhrase_ (Some "Hey Folks!")
     *> Freya.Optic.map Response.body_ (fun body -> body.Write (m, 0, m.Length); body)
```

That throws an exception, because what the lens based modification (lens based map) is doing here is getting the body, writing to it, then returning the body - i.e. it's getting the body from the OWIN dictionary, then setting it on the OWIN dictionary. However:

https://github.com/SuaveIO/suave/blob/master/src/Suave/Owin.fs#L611

Setting the response stream is explicitly disallowed in Suave. Ok, this will work if I change it to be this simple get plus mutational write...

```fsharp
open System.IO

let m =
    "Hello World!"B

let ok =
        Freya.Optic.set Response.reasonPhrase_ (Some "Hey Folks!")
     *> ((fun (body: Stream) -> body.Write (m, 0, m.Length)) <!> Freya.Optic.get Response.body_)
```

Ok, no errors now. So, trying Freya.Machine...

Now we get errors about setting headers because:

https://github.com/SuaveIO/suave/blob/master/src/Suave/Owin.fs#L603

Again, we're doing the same thing as part of our lens operation - things in the middle of the chain are get, then set (even for a get operation, not just map). This time however, we can't really change this, as the lens based modification to state is pretty fundamental to the way Freya is built and works :(

Now, this works in Katana (and IIS I imagine, though I've never tried it!) I'm not sure what the expected behaviour is (and the OWIN spec isn't really explicit from what I can see).

However, I can see why as a server author, we wouldn't want random new values being set for those things - you expect them to be modified *by reference* fundamentally.

So, bit of an issue for Freya. Here's my suggestion then... Would it be possible to loosen that setting check in the code to only fail if the value that's being set *is not reference equal* to the existing value (and therefore the set is pretty much a no-op, and you could either just do it, or ignore it)?

I am the first to admit that's a bit hacky, but it doesn't fundamentally break how the server works that I can see, and also (I think) will get Freya a lot closer to working (no guarantees, as can't test yet!)

If you want me to submit a pull request for that change I probably can, but thought I'd mention it here first for discussion/violent rejection :)

haf:
[...]

PR #358, opened 2015-12-12
reference equal check for setting some aspects of the owin dictionary…

kolektiv:
… (allowing set when the value means an effective no-op)

See #357 for details of motivation. I've not been able to do an end to end test as getting the full build running is being odd for me on this VM, so a local merge and test is probably sensible! I also am not quite sure of the usage of https://github.com/kolektiv/suave/blob/master/src/Suave/Owin.fs#L646 so I've left that as is - do you think it needs modification in light of the other changes? Perhaps not.


kolektiv:
Do any of your PR builds produce nuget packages that I could test as an extra step?

PR #362, opened 2015-12-18
fix headers parsing + update form example

yuyoyuppe:
Headers were read in ASCII which created i18n problems.

### DIFF
--- a/src/Suave/Web.fs
+++ b/src/Suave/Web.fs
@@ -165,7 +165,7 @@ module internal ParsingAndControl =
         Array.blit a.Array a.Offset buf.Array (buf.Offset + !offset) count
         offset := !offset + count
       })
-    let result = ASCII.toStringAtOffset buf.Array buf.Offset count
+    let result = UTF8.toStringAtOffset buf.Array buf.Offset count
     return result, connection
   }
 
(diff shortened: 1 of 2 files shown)


src/Suave/Web.fs
```diff
@@ -165,7 +165,7 @@ module internal ParsingAndControl =
         Array.blit a.Array a.Offset buf.Array (buf.Offset + !offset) count
         offset := !offset + count
       })
-    let result = ASCII.toStringAtOffset buf.Array buf.Offset count
+    let result = UTF8.toStringAtOffset buf.Array buf.Offset count
```
haf: Is this RFC compliant? Isn't the data always ascii?

haf:
Thanks, fixing this would be very good. I created the example you are referring to in order to expose the issue. I'm not sure what is the recommended way to fix it is though. Headers are for example, always ASCII.

ademar:
When I researched this I got the impression that it was no so simple. As Henrik says this should be encoded in ascii.

yuyoyuppe:
Sorry, let me explain my thinking.

>Headers are for example, always ASCII.
Yes, but what are the actual repercussions of using UTF8 instead since it's backward-compatible with ASCII? 

Regarding [multiform-data encoding](http://stackoverflow.com/a/17818574/657390):
It appears that with no charset specified the default is UTF8, so maybe instead of the initial fix, we can make Web.ParsingAndControl.parseMultipart function aware of *accept-charset*/*_charset_* attributes for encoding detection. However, without it specified, the default is UTF8 again.

haf:
Well, perhaps it is uri encoded? To find some authority on how the other servers do it and how the RFCs say it should be done is the first step. 

The other repercussion may be if we have a non ASCII string and want to write it to a header, for example. Then we will corrupt that header and any parser. 

This is similar to how Python 2 made strings be both bytes and chars, and in v3 were "forced" to deprecated their string as byte handling.

PR #367, opened 2016-01-04
listenTimeout

sergey-tihon:
Not sure about it, but looks like `timeout` was renamed to `listenTimeout`


haf:
@sergey-tihon Thank you!

Issue #370, opened 2016-01-06
HTTP request headers search in lower case


kos59125:
Request headers are stored in lower case while `HttpRequest.header` function doesn't take care of letter case. If I search a header by string in upper case, it always falls into `Choice2Of2`.

An illustrating code is the following:

```fsharp
match req.header "X-Forwarded-For" with
| Choice1Of2 value -> // never
| Choice2Of2 error -> // always

match req.header "x-forwarded-for" with
| Choice1Of2 value -> // possible
| Choice2Of2 error -> // possible
```

It seems better if the argument of `req.header` is forced to be lower case in the `header` function rather than in user codes.

haf:
I'd say this is a breaking change, but not a big one... What do you think @ademar ? v1 or v2?

haf:
@kos59125 This would be a good PR.

ademar:
Yeah actually surprised it was not already like that.

On Wednesday, 6 January 2016, Henrik Feldt <notifications@github.com> wrote:

> @kos59125 <https://github.com/kos59125> This would be a good PR.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/SuaveIO/suave/issues/370#issuecomment-169374581>.
>


-- 
Ademar Gonzalez
+1-647-891-3606
http://ademar.name
https://github.com/ademar

PR #371, opened 2016-01-08
ignore letter case in request header names

kos59125:
This PR fixes a problem that a query always falls into `Choice2Of2` (failure) when a key to look up contains any character in upper case (#370).

According to RCF 2616, field names of HTTP header are case-insensitive.
Thus, the behavior above is a bug.

How to resolve:
* adding a new function `getFirstIgnoreCase` which searches a key in a collection case-insensitively.
* changing `HttpRequest.header` function to use the `getFirstIgnoreCase` function instead of `getFirst` to follow the RFC.

### DIFF
--- a/src/Suave.Tests/HttpRequestHeaders.fs
+++ b/src/Suave.Tests/HttpRequestHeaders.fs
@@ -7,7 +7,7 @@ open Suave
 [<Tests>]
 let headers (_:SuaveConfig) =
   testList "Request header letter case" [
-    testCase "compare header names case-insentively" <| fun _ ->
+    testCase "compare header names case-insensitively" <| fun _ ->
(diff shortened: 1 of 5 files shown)


haf:
Could you please use https://github.com/SuaveIO/suave/blob/master/paket-files/haf/YoLo/YoLo.fs#L198 ?

haf:
Looking good btw!

kos59125:
I did it.

haf:
Finally, could we have a unit test, do you think?

kos59125:
IKR. I will.

kos59125:
I made a very simple unit test.

haf:
@kos59125 Thank you for your contribution! We'd be delighted if you would send more PRs like this!

PR #373, opened 2016-01-14
Update composing.md

ImaginaryDevelopment:
adding updated example of basic authentication


haf:
Sorry, we should remove the gh-pages branch.

PR #374, opened 2016-01-14
Update composing.md

ImaginaryDevelopment:
updating another basicAuth section, not sure which goes to the actual site, but both are out of date


haf:
Thank you @ImaginaryDevelopment That's great!

PR #377, opened 2016-01-17
Fix routing docs

Martin4ndersen:
https://suave.io/routing.html

The code results in the error "Expecting a type supporting the operator
'>=>' but given a function type. You may be missing an argument to a
function." due to the lack of `open Suave.Operators`. This is caused by
the need to open `Suave.Operators` to get access to the "fish" operator in
release 1.0.0.


haf:
@Martin4ndersen Thank you!

Issue #379, opened 2016-01-18
Buffer underflow with several parallel requests


simendsjo:
Suave 1.0 from NuGet on .NET 4.5.0 and F# 4.0

I found an issue when adding a parallel testcase in my project. This is only triggered when I spawn many parallel requests (30+). I don't have time to debug at the moment, but I'm adding the issue here in case it's a known issue. `Suave.ParsingAndControl.readMoreData` is called when the buffer is empty (is my first guess).

```
[W] 2016-01-18T16:01:04.6927634Z: tcp request processing failed [Suave.Tcp.job] exn:
System.InvalidOperationException: Stack empty.
   at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
   at System.Collections.Generic.Stack`1.Pop()
   at Suave.Sockets.BufferManager.PopBuffer(FSharpOption`1 context) in Z:\ademar\Projects\suave\src\Suave\Sockets\BufferManager.fs:line 26
   at Suave.ParsingAndControl.readMoreData@113.Invoke(Unit unitVar) in Z:\ademar\Projects\suave\src\Suave\ParsingAndControl.fs:line 113
   at Microsoft.FSharp.Control.AsyncBuilderImpl.callA@851.Invoke(AsyncParams`1 args)
```

EDIT: I'll try to create a smaller testcase in the following days if necessary.

haf:
Might be a duplicate of #363

simendsjo:
Increasing `maxOps` as noted in #363 makes the buffer underflow issue disappear. Seems like a hack though.

haf:
It's not a hack, but how computers work with finite resources. ;) 

The fix [is easy](https://github.com/SuaveIO/suave/issues/363#issuecomment-172457140) – if you're going to use this in production, consider spending an hour to solve it? We'd be grateful for your contribution.

simendsjo:
I was thinking about me just increasing `maxOps` as the hack. I don't know the number of concurrent clients, so I'll just end up pushing this too far up. I'll see if I can take a stab at it during the weekend.

ronnieholm:
@simendsjo: You can override the default Suave config to make Suave log the number of concurrent requests and how much buffer space is left:

    startWebServer { defaultConfig with maxOps = 10; logger = Logging.Loggers.saneDefaultsFor Logging.LogLevel.Verbose } webPart

The buffer space isn't that large: 827392 bytes for the default maxOps of 100.

[...]

Issue #363, opened 2015-12-23
Auto-expanding BufferManager needed


lefthandedgoat:
I was doing some learning on the Http.fs 2.X and I kept crashing my suave server.

Server:
```
let webPart =
  choose
    [
      path "/" >>= choose [ GET >>= (OK "hello world") ]
    ]

startWebServer defaultConfig webPart
```

Crashing code:
```
open HttpFs.Client //this is the pre release v2.4 nuget
open System

let doit () =
    let request = createRequest Get (Uri("http://localhost:8083"))
    let response = getResponse request |> Async.RunSynchronously
    response.StatusCode
    
[| 1 .. 2000 |]
|> Array.iter (fun i -> printfn "%i %A" i (doit()))

System.Console.ReadKey() |> ignore
```

Its crashing because I am doing `let response` instead of `use response` so connections are being left open, but at the 50th connection attempt the exe hosting my suave server will crash.

On OSX+mono the app making the http requests and not closing them just hangs after 3-4 requests, but the suave server is still responding via a browser.

![image](https://cloud.githubusercontent.com/assets/725225/11980463/94fe191c-a960-11e5-8738-7c7c6b40803d.png)

lefthandedgoat:
I don't know if its the same issue, but I was also able to crash it on OSX+Mono by simply issuing a whole bunch of concurrent requests.  I used the osx proxy called charles. Its like fiddler and will record your requests and you can repeat them.  It has an 'advanced repeat' option that allows you to issue X total requests with Y concurrency.  I figure anything over threadpool size of concurrency will push the suave server too hard.  Below is the stack I received.
```
Unhandled Exception:
System.InvalidOperationException: Stack empty.
  at System.ThrowHelper.ThrowInvalidOperationException (ExceptionResource resource) <0x47cd4a8 + 0x0002f> in <filename unknown>:0 
  at System.Collections.Generic.Stack`1[T].Pop () <0x2badb48 + 0x0001f> in <filename unknown>:0 
  at Suave.Sockets.BufferManager.PopBuffer (Microsoft.FSharp.Core.FSharpOption`1 context) <0x2bad980 + 0x0003b> in <filename unknown>:0 
  at Suave.Tcp+job@108.Invoke (Microsoft.FSharp.Core.Unit unitVar) <0x2bad3d8 + 0x001fb> in <filename unknown>:0 
[...]

PR #380, opened 2016-01-19
auto expanding buffer manager

ademar:
ref #379 
ref #363


haf:
Adding a property to a F# Record is a breaking change. Also, adding a parameter to the `create` function is a breaking change. Either we remove those breaking changes, do them another way that's not breaking or we bump the major version.

haf:
+1 on the functionality :)

ademar:
There are more (possibly breaking) changes coming.

haf:
For the c'tors you can create overloads.

I was thinking we can have a 'v2module' module that has globals for things like `autoGrow` in the configuration & comment out the new configuration field.

When we have half a year worth of these things, we just release v2 and remove the `v2module`

ademar:
Actually I was thinking of scratching the configuration field `autoGrow`, at least for now, and merge these changes under v1 with `autoGrow` enabled.

What do you think?

haf:
I think autoGrow should be on by default, but I still think we should not break public APIs like the c'tors.

src/Suave/Sockets/TcpTransport.fs
```diff
@@ -1,38 +1,43 @@
...
 type TcpTransport(acceptArgs     : SocketAsyncEventArgs,
-                  acceptArgsPool : ConcurrentPool<SocketAsyncEventArgs>,
-                  readArgsPool   : ConcurrentPool<SocketAsyncEventArgs>,
-                  writeArgsPool  : ConcurrentPool<SocketAsyncEventArgs>) =
-  let socket = acceptArgs.AcceptSocket
-  let readArgs = readArgsPool.Pop()
-  let writeArgs = writeArgsPool.Pop()
+                  readArgs     : SocketAsyncEventArgs,
+                  writeArgs     : SocketAsyncEventArgs,
+                  transportPool : ConcurrentPool<TcpTransport>,
```
haf: So this is also breaking; can we do something about it?

src/Suave/Tcp.fs
```diff
@@ -32,39 +32,40 @@ let stopTcp (logger : Logger) reason (socket : Socket) =
...
+  acceptArg.add_Completed(fun a b -> userToken.Continuation b)
 
-    readAsyncArgsPool.Push readEventArg
+  new TcpTransport(acceptArg,readEventArg,writeEventArg, transportPool,listenSocket)
 
-    let writeEventArg = new SocketAsyncEventArgs()
-    let userToken = new AsyncUserToken()
-    writeEventArg.UserToken <- userToken
-    writeEventArg.add_Completed(fun a b -> userToken.Continuation b)
+let createPools listenSocket logger maxOps bufferSize autoGrow =
```
haf: Has changed signature; was this function not public?

src/Suave/WebSocket.fs
```diff
@@ -1,217 +1,217 @@
...
-            do ()
-          | Choice2Of2 err ->
-            Log.log ctx.runtime.logger "Suave.Websocket.handShake" LogLevel.Info (sprintf "websocket disconnected: %A" err)
-          return! Control.CLOSE ctx
-        | _ ->
-          return! RequestErrors.BAD_REQUEST "Missing 'sec-websocket-key' header" ctx
-      | _ ->
-        return! RequestErrors.BAD_REQUEST "Bad Request" ctx
-    }
+namespace Suave
```
haf: Can we remove the whitespace changes here?

ademar:
These entities/functions are not intended for public consumption, they are implementation details, we should perhaps mark them internal.  I don't expect anybody would be using any of these anyways.

I have more breaking changes queued towards reducing the number of object allocations on request time. so maybe we could start a branch v2.x with all these changes. or just up the version in master?

What would you do ?

haf:
Up version of master and keep a bugfix branch `releases/v1.x`. Work slightly more with milestones.

ademar:
> Up version of master and keep a bugfix branch releases/v1.x. Work slightly more with milestones.

OK, I'll do that and backport the autogrow logic to v1.1 without modifying the constructors.

Issue #360, opened 2015-12-16
Errors under load for LibUvServerFactory


hodzanassredin:
Trying to do load tests with Yandex.Tank rps_schedule=line(100,40000,2m)
with DefaultTcpServerFactory everything is ok and no errors at all. Ubuntu vm can handle to 4k rps.
But when I try to use LibUvServerFactory exceptions start to occur when load is more than 1400 rps and virtual users more than 31.
<img width="732" alt="benchs" src="https://cloud.githubusercontent.com/assets/40375/11836067/2846eea6-a3e9-11e5-8881-a794d2f09b79.png">

I need to note that kestrel which also uses libuv also does'n work well, it doesnt show errors and works but it cant handle more than 200 rps on the same machine. So probably there is a problem in config.
I installed libuv as described here https://docs.asp.net/en/latest/getting-started/installing-on-linux.html
I tested both suave versions from nuget and from master branch. 

```
let port = 80
let conf  = { defaultConfig with bindings = [ HttpBinding.mkSimple HTTP "0.0.0.0" port ];
                             tcpServerFactory = LibUvServerFactory()
                             bufferSize            = 2048
                             maxOps                = 100}


startWebServer conf (OK "Hello Suave")
```

```
[W] 2015-12-16T08:13:47.5331920Z: tcp request processing failed [Suave.Tcp.job] exn:
System.InvalidOperationException: Stack empty.
  at System.ThrowHelper.ThrowInvalidOperationException (ExceptionResource resource) <0x41a57190 + 0x00027> in <filename unknown>:0
  at System.Collections.Generic.Stack`1[T].Pop () <0x41a06040 + 0x0001f> in <filename unknown>:0
  at Suave.Sockets.BufferManager.PopBuffer (Microsoft.FSharp.Core.FSharpOption`1 context) <0x41a059a0 + 0x0005b> in <filename unknown>:0
  at Suave.Web+ParsingAndControl+readMoreData@115.Invoke (Microsoft.FSharp.Core.Unit unitVar) <0x41a11830 + 0x00087> in <filename unknown>:0
  at Microsoft.FSharp.Control.AsyncBuilderImpl+callA@851[b,a].Invoke (Microsoft.FSharp.Control.AsyncParams`1 args) <0x419f9510 + 0x0015a> in <filename unknown>:0

Unhandled Exception:
System.InvalidOperationException: Stack empty.
[...]

PR #383, opened 2016-01-29
Performance enhancements (implements buffered output)

ademar:
Minimize the number of `send` operations and buffer churning by buffering output till the buffer is full or `flush` is called.

With these set of changes performance/latency comparable to node.js is achieved.

### DIFF
--- a/examples/Example/CounterDemo.fs
+++ b/examples/Example/CounterDemo.fs
@@ -47,4 +47,5 @@ let counterDemo (req : HttpRequest) (out : Connection) =
         (lastEvtId - 1u)
 
     for a in actions do
-      do! a }
+      do! a
+    return out }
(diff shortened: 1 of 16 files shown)


haf:
:cake:

scitesy:
I am curious. Is the performance/latency equivalence to node.js with or without a  libuv dependency?

ademar:
@scitesy Hi Scott. This is with the default tcp engine.  At the moment the libuv implementation seems to run slower.

scitesy:
This is an exciting development! This work seems to address some of the questions raised in issue #360?

ademar:
> This work seems to address some of the questions raised in issue #360?

Yes.

scitesy:
Excellent! :) Thank you.

Issue #398, opened 2016-02-26
OWIN dictionary keys not compared using Ordinal string comparison


Andrew-Webb:
The [OWIN specification](http://owin.org/html/spec/owin-1.0.html) is clear:-

* the keys of the environment IDictionary&lt;string, object&gt; must be compared using `StringComparer.Ordinal` (i.e. non-cultural, case-sensitive)
* the keys of the contained headers IDictionary&lt;string, string[]&gt; must be compared using `StringComparer.OrdinalIgnoreCase` (i.e. non-cultural, case-insensitive)

Suave's OWIN implementation uses case-insensitive comparison (+ equality test) for both types of dictionary, so it's half right and half wrong.

The following illustrates:-

    // Works, but shouldn't...
    let headers = state.Environment.["owin.requestheaders"] :?> IDictionary<string, string[]>

    // Works, and should
    let host = headers.["HOST"]     // the key in the dictionary is actually "host"

Andrew-Webb:
@ademar: Thank you for addressing this issue, as well as #397 and #387 - great work!
I look forward to trying it.

To me it makes sense to create the underlying .NET dictionaries directly, specifying the correct `StringComparer` value in the ctor of these dictionaries.

However, there is a residual use of `StringComparer.OrdinalIgnoreCase` outside of dictionary creation [on this line](https://github.com/SuaveIO/suave/blob/master/src/Suave/Owin.fs#L344).

I don't know for sure - I've just perused the code - but it seems to me that if `mapFindLens` were being used for the headers dictionary then this would make sense, but I believe it's being used for the environment dictionary, in which case maybe `StringComparer.Ordinal` would make more sense?

Just my two cents after a short and cursory "code review".

ademar:
@statpro-awebb thanks! it is fixed now 3d317e92a90c4bdac1e4266795c87763119f0c28

panesofglass:
@Andrew-Webb, wow, as one of the OWIN maintainers, I never realized the environment dictionary keys were `StringComparison.Ordinal`. Thanks for the catch!

Thanks for fixing this, @ademar!

Issue #390, opened 2016-02-17
Custom HTTP status codes


simendsjo:
I'm using Suave in an internal application. I would like to use custom statuscodes to indicate some warnings in the application, but unfortunately, Suave only supports standard HTTP status codes. This means I would have to reuse a statuscode to indicate a different condition specific to my application.

Would a PR to allow custom statuscodes be accepted? Or is the stance only to allow these predefined status codes? As Suave is a good fit for microservices that has an HTTP API, I would hope for the former.

haf:
Yes, but not if you change any code that concerns HTTP status codes. Instead, let me point you towards how it could be done nicely. Suave is built in a modular fashion, so if you set your HttpResult to *not* write the [preamble](https://github.com/SuaveIO/suave/blob/master/src/Suave/ParsingAndControl.fs#L403-L459) you can replace it with your own status code.

You can create a combinator that replaced the preamble with a custom HTTP code. That should make your WebPart just a few lines of code, while not removing the totality of functions with HttpCode inputs.

What do you think?

PR #399, opened 2016-02-27
Custom http status codes

### LINKED ISSUE #387: OWIN Dictionary TryGetValue only works for known keys?
It seems from this, at first sight, that https://github.com/SuaveIO/suave/blob/master/src/Suave/Owin.fs#L712-L717 means that TryGetValue will only return true and a value if the key was in a pre-existing set of keys. This means: 

```fsharp
dict.["someKey"] <- "hello"

let hello = dict.["someKey"]

// hello = "hello"

let maybeHello =
    match dict.TryGetValue "someKey" with
    | true, v -> Some v
    | _ -> None

// maybeHello = None (!)
```

Freya adds some values in to the OWIN dictionary as it processes requests, but given this, it can't get them out again using TryGetValue (which is how the relevant Freya optic for environment dictionaries works)!

Is this intended behaviour? I would expect that the OWIN dictionary would, to external consumers, act like a normal dictionary...

### LINKED ISSUE #397: Reason phrase should be settable via OWIN
Arising from discussion in #387 

Although use of the reason phrase is downplayed by [RFC 7230](http://tools.ietf.org/html/rfc7230#section-3.1.2), there seems to be no good reason for not allowing it to be set in Suave via OWIN.

ademar:
This PR fixes #387, #397, #390 and #398


Gemfile.lock
```diff
@@ -21,6 +21,3 @@ PLATFORMS
 
 DEPENDENCIES
   albacore (~> 2.3)
-
```
haf: `gem update bundler` to get latest

PR #404, opened 2016-03-12
rewrite HTML to be more aligned with spec

### LINKED ISSUE #403: Suave.Html should be updated to follow the HTML spec
Suave.Html is assuming the content is `XML` and not `HTML`, and it is a slight difference. In `HTML` self-closing tags are not permitted, the are referred to as [void elements](https://dev.w3.org/html5/html-author/#void). 

The consequence of this is that the correct `XML` is generated, but that is not interpreted by the browser as expected. The easiest example is the `script` tag with the `src` attribute set. This will not be interpreted as expected by the browsers as the implementation is today.

I actually think the solution as it is now should be rewritten to target `HTML` and generic `XML`.

mastoj:
This should fix #403. I kept renamed the old one to Suave.Xml, since that is more generic and not true HTML as by the specification.

I also simplified the code for the HTML part.

The renaming is a breaking change for the HTML, but thought it is ok since it is in Experimental. Ok?

### DIFF
--- a/src/Experimental/Form.fs
+++ b/src/Experimental/Form.fs
@@ -38,15 +38,15 @@ type Form<'a> = Form of FormProp<'a> list * ServerSideValidation<'a> list
 
 let formatDec (d : Decimal) = d.ToString(Globalization.CultureInfo.InvariantCulture)
 
-let (|Optional|_|) (typ : Type) = 
-  if typ.IsGenericType 
-     && typ.GetGenericTypeDefinition() = typedefof<option<_>> then 
+let (|Optional|_|) (typ : Type) =
+  if typ.IsGenericType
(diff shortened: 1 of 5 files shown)


ademar:
Thanks Tomas!

mastoj:
No problem. I also think the second parameter for the `Element` type could be removed. I kept it since it was in the original implementation, but I can't figure out what it is good for.

```
type Element = string * Attributes[]
```

should be enough.

ademar:
It made sense in the `XHTML` implementation:

https://github.com/SuaveIO/suave/blob/master/src/Experimental/Xml.fs#L10

Note that code in the Experimental folder is just stuff and ideas I started but never got around to complete.

PR #417, opened 2016-03-28
Custom mapJson2

mastoj:
The existing mapJson only allows for the standard serializer.
mapJson2 (name can be discussed) makes it possible to provide
serializer and deserializer.


mastoj:
Maybe not something you have asked, but it makes it easier to plugin any serializer and deserializer. I'm not satisified with the name of the new function, if you think this is useful we could come up with a better name.

ademar:
What about calling it `mapJsonWith` ?

mastoj:
@ademar good suggestion :). Updated the name. In a more advanced scenario you could imagine an even more generic function that also take care of other formats then json, but that might be something for the future.

PR #421, opened 2016-04-03
basic authentication with cookie suppport

mastoj:
When using basic auth the header isn't accessible that easy from javascript, or at least I didn't find a nice way. Storing the authorization data in a cookie when logged in makes it easy to create a web page with js-app that supports basic auth, the js-app can just use the cookie.

If rejected, please tell how it should be done. If accepted, please check and comment on the `authenticateBasicWithCookie` method. It feels like it might be a nicer way to write it.

### DIFF
--- a/src/Suave/Authentication.fs
+++ b/src/Suave/Authentication.fs
@@ -84,6 +84,37 @@ let authenticate relativeExpiry secure
       decryptionFailure
       fSuccess)
 
+let internal validateAuthHeader validationFunction (ctx:HttpContext) =
+  let req = ctx.request
+  match req.header "authorization" with
+    | Choice1Of2 header ->
+      let (typ, username, password) = parseAuthenticationToken header
+      if (typ.Equals("basic")) && validationFunction (username,password) then
+        ASCII.bytes header |> Choice1Of2
+      else
+        challenge |> Choice2Of2
+    | Choice2Of2 _ ->
+      challenge |> Choice2Of2
+
+let authenticateBasicWithCookie relativeExpiry secure validationFunction (protectedPart:WebPart) =
+  let continuation =
+    context(fun ctx ->
+      match ctx.response.cookies |> readCookies ctx.runtime.serverKey SessionAuthCookie with
+      | Choice1Of2 (cookie, cookieValueBytes) ->
+        cookieValueBytes
+        |> ASCII.toString
+        |> parseAuthenticationToken
+        |> (fun(_,userName,_) -> addUserName userName ctx)
+        |> (fun ctx' -> (fun _ -> protectedPart ctx'))
+      | Choice2Of2 _ -> challenge)
+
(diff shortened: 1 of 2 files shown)


ademar:
Hi Tomas,

Yeah I'm kind of on the fence about this. Would be nice to get also @haf's opinion.

The thing is HTTP Simple Authentication have nothing to do with cookies. 

> HTTP Basic authentication (BA) implementation is the simplest technique for enforcing access controls to web resources because it doesn't require cookies, session identifiers, or login pages

And I wouldn't want to pollute our library with made-up authentication schemes.

> If rejected, please tell how it should be done. 

The simplest would be just to include this code in your own project.

Cheers,

ademar

haf:
I use the Authorization header all the time with https://github.com/logibit/logibit.hawk/tree/master/src/logibit.hawk – admittedly not *basic* authorization. The function on `XMLHttpRequest` is called `setRequestHeader`, and the third and fourth arguments of `open` are also for Basic auth, I believe (but I've never used them as they're spotty across browsers).

With this technique you're exposing your users' passwords to anything that can read cookies on the computer and by doing so you're weakening the client's security. I would not want internet apps that I use to save my credentials in cookies. If you derive an API key from the password and keep that secure and then store it in something, like local storage that's a bit better, since you're not sending your key in every request and the key is not your password.

mastoj:
I think I agree, that this should probably not be merged. 

I know I can set the header from a js-webapp, but I have that means I might have to login twice if I have a standard route that I have "protected" and the js is protected since the javascript can't read the header from originate request. Javascript can read headers if the javascript initiate the request.

Not sure I think the difference between sending the credentials in every request with a cookie or a header is such a big difference though, but I do agree that this opens up for users to do the "wrong" thing. I was just doing this for a demo, and thought it might be useful.

haf:
There's no big difference in the medium over which you send credentials, like you state. My point was that it's better to use a derived key rather than the user secret, which may be re-used across sites by the user himself/herself. The secondary point was that you can use a proof of having the secret, like HMAC on the client rather than sending the secret over and over, which stops a lot of indirection attacks like ARP poisioning, DNS poisioning, HTTPS downgrade and unsanitised input redirection attacks.

Can't you send a body to the JS that authenticates? http://docs.qvitoo.apiary.io/#reference/accounts/authenticate/authenticate

mastoj:
What you suggest is definitely something one should do in a "real" application and I totally agree with all that you say. For this demo I think I'll just go with forms authentication to keep the demo simple I think.

A side note: Great work with Suave! I really like what you done! Both of you.

haf:
DHCP highjacking kicked my ass today...

On a side note, I think consensus is not to merge this, so I'm closing the PR. None the less, we do very much appreciate you being here and trying to improve Suave with us. Keep that spirit up! =)

PR #432, opened 2016-05-02
Update paket.targets to latest version

danielrbradley:
Fixes building when the paket.references path has a space in it.

Sourced from the [latest release in the paket repo](https://github.com/fsprojects/Paket/blob/22e268cd65463d30cc5f501047d903170c33c4f5/.paket/paket.targets)


ademar:
Thank you!

PR #433, opened 2016-05-04
Add mimetype application/woff2 used by FontAwesome (and others)

ericharding:
A tiny change to add the woff2 mime type.  I ran into this when trying to serve FontAwesome with Suave.  I believe the file format is still in draft but it seems like it's getting used.

https://dev.w3.org/webfonts/WOFF2/spec/
http://stackoverflow.com/questions/28235550/proper-mime-type-for-woff2-fonts


ademar:
Thank you!

PR #436, opened 2016-05-14
Add link to Music Store tutorial

ovatsus:
I had never heard of the tutorial before, and it's great, so it really should be linked from the Suave website so newcomers know about it


ademar:
Thanks!

PR #441, opened 2016-05-22
Create azure-app-service.md

isaacabraham:
First draft of (simple) documentation to deploy and host Suave in the Azure App Service.

### DIFF
--- a/docs/azure-app-service.md
+++ b/docs/azure-app-service.md
@@ -1,0 +1,55 @@
+---
+layout: default
+---
+
+Deploying Suave to Azure App Service
+------------------------------------
+
+Suave Web Sites in Azure App Service can be deployed either as an F# script with a suitable host e.g. FAKE, or as a standalone executable. This tutorial will focus on: -
+
+* Deploying and building your Suave application from source control directly in the Azure App Service  
+* Hosting a Suave console application in Azure App Service
+* Instructing the Azure App Service IIS host to redirect all traffic to Suave
+
+### 1. Create a ``web.config`` file.
+This file instructs IIS to act as a passthrough and to redirect all traffic through to Suave.
+
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+  <system.webServer>
+    <handlers>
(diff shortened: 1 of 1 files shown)


ronnieholm:
Is automatically deploying on every commit a good practice in real-world scenarios? It seems all too easy to introduce a bug in production. I FTP-deploy my Suave app to Azure, but it's a bit of hazzle. Mainly because I have to stop the App Service every time, deploy, and then restart.

Would you have a separate dev branch and then only automatically deploy on commits to master?

isaacabraham:
@ronnieholm That's definitely one way to do it i.e. have a specific branch for release. I've done that on several systems, it works fine - but I've also been looking at other options such as Octopus / Appveyor / VSTS in order to separate out releases from source code management.

I'm not sure that getting into a discussion of that is really appropriate here though - the crux is the web.config + .deployment files; the git stuff is just a means to an end to get the executable on the box (as you rightly point out, FTP is another way).

As a side note, you shouldn't need to stop / start / restart your site - you should really use deployment slots on the service. Deploy to a secondary slot, make sure you're happy with it, then flip the slots and you won't need to worry about stop / start / restart. This is (again) a concern which IMHO is sort of orthogonal to Suave though :-)

ronnieholm:
Sorry. I didn't mean to start a specific/othogonal discussion. My concern was that users reading the guide would feel stuck once they move off their toy app and onto a real app. Personally, when I moved away from VS deployment, I felt the Azure (again, not specific to Suave) deployment story very confusing. Hence I defaulted to FTP.

isaacabraham:
I agree - the documentation across all of Azure is *very* much geared up towards a VS-first story. This hurts everyone (even VS users) so I'm glad to see some efforts being made to rectify this. There are some much improved docs [here](https://azure.microsoft.com/en-gb/documentation/articles/web-sites-deploy/) although I'm sure even more could be done to improve things.

ademar:
Thanks!

isaacabraham:
Sorry - Just realised that I didn't update the layouts page.

ademar:
I'll link it there.

PR #445, opened 2016-05-24
Emit preamble for EventSource

ranma42:
EventSource should emit the preamble, but it communicates over the
connection through the asyncWriteBytes primitive, which does not flush
the line buffer.

This can be fixed by manually flushing the buffer before starting the
event data stream.

Fixes the output of examples/Example/es.html

### DIFF
--- a/src/Suave/Combinators.fs
+++ b/src/Suave/Combinators.fs
@@ -697,6 +697,7 @@ module EventSource =
   let private handShakeAux f (out : Connection, _) =
     socket {
       let! (_,out) = asyncWriteLn "" out// newline after headers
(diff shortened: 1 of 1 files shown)


haf:
@ranma42 Thank you!

ranma42:
Any idea about why AppVeyor is failing?

haf:
Yes, because the build is broken on CoreCLR. We're working on it /cc @enricosada 

Here's the build output that's breaking it: https://ci.appveyor.com/project/haf/suave/build/1.0.137/job/tu9qg6xpang9rfl8

enricosada:
i'll send a pr about that now

PR #446, opened 2016-05-24
fix appveyor build

enricosada:


enricosada:
we can use the `*` the final part of the version inside project.json (not for the whole version number, not supported)
i'll send another pr.

PR #450, opened 2016-05-26
Fix code formatting

dsyme:
Fix the code to be a bit more idiomatic


haf:
@dsyme Thank you for your contribution to the docs!

PR #451, opened 2016-05-26
Simplify code for initial readers

dsyme:
I was showing Suave to a F# beginner today and noticed that in some cases the examples were using more advanced or rarer F# idioms in a way that's not necessary to get across the point of what's going on.

For example it is generally better not to use ``<|`` in beginner samples unless specifically explaining that operator. Just use parentheses in beginner code.  Similarly don't use ``((=) (a,b))`` in beginner code.

The code here also contained a mistake with an end-of-line comment and the closing brackets of some lists.


haf:
I agree. Thank you!

Issue #458, opened 2016-06-06
Getting error: Key 'filename' was not present in 'content-disposition'


toburger:
A client is calling my suave server with Content-Disposition: form-data and I'm getting the error from above.

The request I receive is the following:
```
POST http://127.0.0.1:8083/ HTTP/1.1
Content-Type: multipart/form-data; boundary="2fb29e4c-34b4-40d4-b8f4-f6f561e65539"
Host: 127.0.0.1:8083
Content-Length: 342
Expect: 100-continue
Connection: Keep-Alive

--2fb29e4c-34b4-40d4-b8f4-f6f561e65539
Content-Type: text/plain; charset=utf-8
Content-Disposition: form-data; name="action"

OTA_HotelInvNotif:Inventory
--2fb29e4c-34b4-40d4-b8f4-f6f561e65539--
```

and it turns out, that the error gets thrown when the Content-Disposition is used in conjunction with Content-Type. If Content-Type is omitted everything works. I wonder if the client isn't following the HTTP specifications or is it a bug in suaves multipart parser?

ademar:
Suave is expecting a `filename` key in the Content-Disposition header.  It
seems this is not required, so we will need to fix it.

On Mon, Jun 6, 2016 at 9:06 AM, Tobias Burger <notifications@github.com>
wrote:

> A client is calling my suave server with Content-Disposition: form-data
> and I'm getting the error from above.
>
> The request I receive is the following:
>
> POST http://127.0.0.1:8083/ HTTP/1.1
> Content-Type: multipart/form-data; boundary="2fb29e4c-34b4-40d4-b8f4-f6f561e65539"
> Host: 127.0.0.1:8083
> Content-Length: 342
> Expect: 100-continue
> Connection: Keep-Alive
>
> --2fb29e4c-34b4-40d4-b8f4-f6f561e65539
> Content-Type: text/plain; charset=utf-8
> Content-Disposition: form-data; name="action"
>
> OTA_HotelInvNotif:Inventory
> --2fb29e4c-34b4-40d4-b8f4-f6f561e65539--
>
> and it turns out, that the error gets thrown when the Content-Disposition
> is used in conjunction with Content-Type. If Content-Type is omitted
> everything works. I wonder if the client isn't following the HTTP
> specifications or is it a bug in suaves multipart parser?
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
[...]

PR #462, opened 2016-06-08
Fix for #458

toburger:
According to http://www.faqs.org/rfcs/rfc2388.html can provide Content-Disposition without the filename attribute.

I don't know if this is a good solution, because the provided content type gets discarded.
Maybe the content type should be added to the multiPartFields list with a default of text/plain?


ademar:
Fix looks good. I agree it makes sense to collect the content-type.

PR #469, opened 2016-06-16
Logging combinator that allows to set LogLevel

ademar:


haf:
Actually I think it should be using Debug. Throughout my own app, Logary, etc I have it so that Debug means a low-level metric/event and Info is something more worth to my business. Why Debug and not Verbose? Because at Verbose you get too much data. Warn for business events? Sort of doesn't match intention. Both at Info? Maybe, but I'd have to put up pretty draconian namespace filters and that takes time so it's not a good default.

ademar:
OK. I've factored instead a new combinator that let me set LogLevel.

haf:
+1

PR #472, opened 2016-06-20
fix for .net core sdk preview1

enricosada:
Some fixed, based on https://github.com/dotnet/netcorecli-fsc/wiki/.NET-Core-SDK-preview1#notes

This fix a minor mistake of `preview1`, and helps to not break build because works is in progress for `preview2` .

### DIFF
--- a/examples/Example/project.json
+++ b/examples/Example/project.json
@@ -23,7 +23,7 @@
 
   "tools": {
     "dotnet-compile-fsc": {
-      "version": "1.0.0-*",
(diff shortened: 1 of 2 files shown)


enricosada:
NOTE: the existing package is ok. that's only a development thing

i already fixed the https://github.com/SuaveIO/Suave-CoreCLR-sample

ademar:
Thanks!

enricosada:
np @ademar i'll continue to maintain support for .net core in suave, it's new, so fix happen :smile:

PR #476, opened 2016-06-27
add support for array, fixing naming mistakes

jackmott:
This adds support for array along with seq and list.
I also fixed a minor copypasta mistake where "razor" was mentioned, and 
some comment typos.


ademar:
Thanks!

PR #478, opened 2016-06-27
dotliquid registerFiltersByName, options, arrays

jackmott:
This fixes dotLiquid registerFiltersByName, such that it will search the users assembly instead of only the suave.DotLiquid.dll assembly, for modules of the name `name`. Also switched to a probably marginally more efficient search of the assemblies which is no less readable using Array.find

This also fixes array support, types within arrays are now properly registered. This was not working properly in my previous PR/

This also adds support for options, by registering the Value field you can use them in templates like so:

```html
{% for decoder in model.decoders %}  
  {% if decoder.Value %}               
    <option value="{{decoder.Value.x}}">{{decoder.Value.x}}</option>
  {% endif %}
{% endfor %}
```

### DIFF
--- a/src/Suave.DotLiquid/Library.fs
+++ b/src/Suave.DotLiquid/Library.fs
@@ -21,7 +21,6 @@ open Suave.Files
 /// Use the ruby naming convention by default
 do Template.NamingConvention <- RubyNamingConvention()
 
-
 module internal Impl =
 
   /// Represents a local file system relative to the specified 'root'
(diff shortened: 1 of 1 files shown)


jackmott:
I can't tell why the AppVeyor test failed, any tips?

adamchester:
It's the wrong obtain URL now, or perhaps they deleted the old CLI tools from their temporary azure storage.

adamchester:
I would imagine it's best to do this in a separate PR.

jackmott:
Any concerns with this PR? Checks pass now.

ademar:
Thanks!

PR #485, opened 2016-07-07
resolves #484 - adds support for sending ArraySegments of existing buffers

### LINKED ISSUE #484: Allow sending of byte ArraySegment's on websocket
When using the Websocket support in Suave the send method on the Websocket object is of the form:

member Send: opcode: Opcode -> bs: byte[] -> fin: bool-> Async<Choice<unit, Error>>

The problem with the above method is that if I'm using this framework with large big-data style messages the large object heap quickly fills up and I get application OutOfMemoryException's due to the use of the byte array as a parameter. Large byte[] arrays go on the large object heap and cause fragmentation quickly.

I'm proposing adding an overload that takes in an ByteSegment instead allowing the user to reuse byte[] array buffers and avoiding the issue of sending large messages to the client along with other optimisations to avoid allocating additional byte arrays before sending it to the SocketAsyncEventArgs (e.g in the frame function of the Websocket module in particular with the MemoryStream being created to build the frame payload). 

member SendSegment: opcode: Opcode -> bs: ByteSegment -> fin: bool-> Async<Choice<unit, Error>>

I've done the change and it doesn't seem too hard on my laptop with tests passing; just wanting your opinion before I bother to put it up whether this is a feature you also want upstream in the library.

mvkra:
This allows the reuse of existing byte array buffers when sending messages via websockets. Some points to note:

- I have replaced the send member of the Websocket class in Websocket.fs as instructed to on the #484 issue.
- If I use the parameter-less constructor for ArraySegment's the LibUV tests cause seg fault native exceptions. This was under Mono 4.2. Using ArraySegment([||]) instead fixes this but feels a little unnatural and forced from an API perspective.

### DIFF
--- a/examples/WebSocket/Program.fs
+++ b/examples/WebSocket/Program.fs
@@ -24,11 +24,11 @@ let echo (webSocket : WebSocket) =
       match msg with
       | (Text, data, true) ->
         let str = UTF8.toString data
-        do! webSocket.send Text data true
+        do! webSocket.send Text (ArraySegment data) true
       | (Ping, _, _) ->
-        do! webSocket.send Pong [||] true
+        do! webSocket.send Pong (ArraySegment([||])) true
       | (Close, _, _) ->
-        do! webSocket.send Close [||] true
+        do! webSocket.send Close (ArraySegment([||])) true
         loop := false
       | _ -> ()
   }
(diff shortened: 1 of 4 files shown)


ademar:
Good stuff. Thanks.

PR #490, opened 2016-07-16
Dispose file streams when serving files

i-p:
Opened file streams were not disposed when serving static files. This caused handle leaks on Windows.

To reproduce this issue, run Suave with configuration to serve static files.
Open any file served by Suave in a browser (it should have at least 500 bytes, so that response is compressed). Now press and hold key that reloads displayed page for a few seconds.

Then use Handle utility (or Process Explorer) where you can see that there are several opened handles
for the original file and also its compressed representation:
```
Handle v4.0
Copyright (C) 1997-2014 Mark Russinovich
Sysinternals - www.sysinternals.com

   ...  
  5D0: File  (---)   C:\dev\suave\src\Suave\bin\Debug\_temporary_compressed_files\g5kv1sjy.n5n
  6A0: File  (---)   C:\dev\suave\index.html
  6BC: File  (---)   C:\dev\suave\src\Suave\bin\Debug\_temporary_compressed_files\g5kv1sjy.n5n
  6C0: File  (---)   C:\dev\suave\index.html
  6C4: File  (---)   C:\dev\suave\src\Suave\bin\Debug\_temporary_compressed_files\g5kv1sjy.n5n
  7DC: File  (---)   C:\dev\suave\src\Suave\bin\Debug\_temporary_compressed_files\g5kv1sjy.n5n
  7E0: File  (---)   C:\dev\suave\src\Suave\bin\Debug\_temporary_compressed_files\g5kv1sjy.n5n
  7E4: File  (---)   C:\dev\suave\index.html
  7E8: File  (---)   C:\dev\suave\src\Suave\bin\Debug\_temporary_compressed_files\g5kv1sjy.n5n
 
```

After adding Dispose() calls, handles for the original file and its compressed representation are closed correctly.

### DIFF
--- a/src/Suave/Combinators.fs
+++ b/src/Suave/Combinators.fs
@@ -461,21 +461,23 @@ module Files =
       let getFs = fun path -> new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.ReadWrite) :> Stream
       let getLm = fun path -> FileInfo(path).LastWriteTime
       let! (encoding,fs) = Compression.transformStream file getFs getLm compression ctx.runtime.compressionFolder ctx
-
-      match encoding with
-      | Some n ->
-        let! (_,conn) = asyncWriteLn (String.Concat [| "Content-Encoding: "; n.ToString() |]) conn
-        let! (_,conn) = asyncWriteLn (sprintf "Content-Length: %d\r\n" (fs : Stream).Length) conn
-        let! conn = flush conn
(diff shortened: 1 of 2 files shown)


ademar:
Nice catch.

However the change to Compression.fs does not seem entirely correct.  Note the function `compressFile` is already disposing the stream.

Perhaps we should remove the dispose call from there and keep your try-finally change?

i-p:
Yes, that `Dispose` call is not needed any more, so I'll remove it, Thanks for noticing that.

ademar:
Thanks!

PR #493, opened 2016-08-02
Provide nice logging API

haf:
This PR implements the Logary Façade without adding a reference to Logary. Instead it's done via a Github reference to the Facade.fs file. By default the façade provides coloured console logging, but when you deploy this library to production I recommend initialising the Façade with a [Logary backend](https://github.com/logary/logary#using-logary-in-a-library).


haf:
Ready to merge. You want to give it a quick glance @ademar ?

ademar:
Looks good :+1:

PR #495, opened 2016-08-15
Add helper function for getting response status code for a request in Suave.Testing

anton-pt:
Small addition to Suave.Testing. I don't know if this helper function was missing due to an omission or if there is a reason for it but seems useful to me... :-)


ademar:
Thanks!

PR #499, opened 2016-08-30
Enhanced structured log messages

adamchester:
The following are addressed:
 1. Fix buffer manager log message. It was setting a `size` field, with a template that required `{totalBytes}` (10611b4)
 1. Improve the 'listener started' log message structure, instead of formatting useful structured data as strings (via`StartData.ToString()`), we now separate out the parts into individual structured fields. (650e749)
 1. Add a structured log combinator, the same as the existing log combinator functions, only with the structured message format `template {prop}` and a `Map<string,obj>` for name/values to include in the structured log event. (b49afc4)


haf:
Nice, thanks @adamchester – this makes it much nicer to get requests logs out of Suave.

PR #504, opened 2016-09-12
Decode multipart data based on charset

toburger:
I found that the encoding is always ASCII. It should be based on the provided charset.
So I made this PR so at least UTF-8 gets respected.
Feel free to correct my code if it is a bit too weird with all this Choice mumbo-jumbo.


ademar:
Thanks!

PR #509, opened 2016-09-23
Fix docs for startWebServerAsync – yields startup metrics

lapponiandevil:
Booh yah!


haf:
Thanks @lapponiandevil !

PR #511, opened 2016-09-26
Add "frameworks: net46" to Paket specifications in script prefixees

dsyme:
This makes paket dependency resolution MUCH MUCH faster


ademar:
Thanks!

ademar:
@dsyme that was commit number 2000 :)

haf:
🌷

PR #512, opened 2016-09-28
Adding HTTP Status Codes, 306 and 426

tamizhvendan:
While working [on a function](https://github.com/tamizhvendan/Suave.Azure.Functions/blob/d427f9dc474748b3b78c2ae48c9b2cc6e2e9953d/tests/Suave.Azure.Functions.Tests/Response.fs#L15-L23) that maps `HttpStatusCode` from `System.Net.Http` to the `HttpCode` of `Suave.Http`, I got an FsCheck error for the HttpStatusCode `306` and `426` as their is no corresponding Status Code in Suave.

This pull request adds these status codes to `HttpCode`.

Reference: [HttpStatusCode Documentation](https://msdn.microsoft.com/en-us/library/system.net.httpstatuscode(v=vs.110).aspx)

image


ademar:
Thanks!

tamizhvendan:
:-)

PR #515, opened 2016-10-03
Handle Cookies from Owin Closes #514

### LINKED ISSUE #514: IdentityServer3 (Owin) Cookies
So I attempting to get IdentityServer3 (2.5.3) Suave (2.0.0-alpha006) running on Mono (4.6.1). I used the implicit flow for it and I was able to get the the login screen.  However when I attempted to login, I was sent back to the login page.  After some digging I found out the for some reason the Set-Cookie headers weren't being set for the cookies that IdentityServer needs (idsvr.session, idsvr, idsvr.partial, etc).  Only the first cookie was being set, SignInMessage.   This also seems true for any Owin middleware that sets more than one cookie.  I doubled checked on Windows .net 4.6 and the behavior was the same.

```
    type Cookies(next:OwinMiddleware)  =
        inherit OwinMiddleware(next)
        override this.Invoke (context:IOwinContext) : Task =
            async {
                context.Response.Cookies.Append("FirstCookie","FirstValue")
                do! next.Invoke(context) |> Async.AwaitTask
                context.Response.Cookies.Append("SecondCookie","SecondValue")
            } |> Async.StartAsTask :> Task
    let setRandomCookies =
        let app = new AppBuilder()
        let app = app.Use(typeof<Cookies>, [||]) 
        app.Build()
        |> OwinApp.ofAppFunc "/" 
```

image

TLDR;
Suave needs to handle Set-Cookie differently when coming back from OWIN.

Owin.fs
```Fsharp
      let handleResponse  (rhs : Dictionary<string, string[]>) = 
        let handleKey key =
            if key = "Set-Cookie" then 
                rhs.[key] |> Seq.map(fun v -> key,v) |> Seq.toList
            else
                [key, String.concat ", " rhs.[key]]

        (Seq.collect handleKey rhs.Keys |> Seq.toList)
        
      let setResponseHeaders (rhs : Dictionary<string, string[]>) =
        Lens.set respHeaders_ (handleResponse rhs)
```

I can send a pull request if you'd like.

TheAngryByrd:
Let me know if there is anything I need to fix up.

### DIFF
--- a/src/Suave/Owin.fs
+++ b/src/Suave/Owin.fs
@@ -635,13 +635,13 @@ module OwinApp =
         HttpContext.response_ >--> HttpResult.headers_
 
       let handleResponse  (rhs : Dictionary<string, string[]>) = 
-        let handleKey key =
-          if key = "Set-Cookie" then 
-            rhs.[key] |> Seq.map(fun v -> key,v) |> Seq.toList
-          else
+        let handleKey = function
+          | "Set-Cookie" as key -> 
+            rhs.[key] |> Seq.map (fun v -> key,v) |> Seq.toList
+          | _ as key -> 
             [key, String.concat ", " rhs.[key]]
 
-        (Seq.collect handleKey rhs.Keys |> Seq.toList)
+        Seq.collect handleKey rhs.Keys |> Seq.toList
 
       let setResponseHeaders (rhs : Dictionary<string, string[]>) =
         Lens.set respHeaders_ (handleResponse rhs)
--- a/src/Suave/Owin.fs
+++ b/src/Suave/Owin.fs
@@ -636,7 +636,7 @@ module OwinApp =
 
       let handleResponse  (rhs : Dictionary<string, string[]>) = 
         let handleKey = function
-          | "Set-Cookie" as key -> 
+          | Headers.Fields.Response.setCookie as key -> 
             rhs.[key] |> Seq.map (fun v -> key,v) |> Seq.toList
           | _ as key -> 
             [key, String.concat ", " rhs.[key]]
--- a/src/Suave/Owin.fs
+++ b/src/Suave/Owin.fs
@@ -634,8 +634,17 @@ module OwinApp =
         HttpContext.request_ >--> HttpRequest.headers_,
         HttpContext.response_ >--> HttpResult.headers_
 
+      let handleResponse  (rhs : Dictionary<string, string[]>) = 
+        let handleKey key =
+          if key = "Set-Cookie" then 
+            rhs.[key] |> Seq.map(fun v -> key,v) |> Seq.toList
+          else
+            [key, String.concat ", " rhs.[key]]
+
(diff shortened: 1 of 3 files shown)


src/Suave/Owin.fs
```diff
@@ -634,8 +634,17 @@ module OwinApp =
...
         HttpContext.response_ >--> HttpResult.headers_
 
+      let handleResponse  (rhs : Dictionary<string, string[]>) = 
+        let handleKey key =
+          if key = "Set-Cookie" then 
+            rhs.[key] |> Seq.map(fun v -> key,v) |> Seq.toList
+          else
+            [key, String.concat ", " rhs.[key]]
+
+        (Seq.collect handleKey rhs.Keys |> Seq.toList)
```
haf: Could remove extra paren

src/Suave/Owin.fs
```diff
@@ -634,8 +634,17 @@ module OwinApp =
         HttpContext.request_ >--> HttpRequest.headers_,
         HttpContext.response_ >--> HttpResult.headers_
 
+      let handleResponse  (rhs : Dictionary<string, string[]>) = 
+        let handleKey key =
+          if key = "Set-Cookie" then 
+            rhs.[key] |> Seq.map(fun v -> key,v) |> Seq.toList
```
haf: Space after map

src/Suave/Owin.fs
```diff
@@ -634,8 +634,17 @@ module OwinApp =
         HttpContext.request_ >--> HttpRequest.headers_,
         HttpContext.response_ >--> HttpResult.headers_
 
+      let handleResponse  (rhs : Dictionary<string, string[]>) = 
+        let handleKey key =
```
haf: What's your opinion on

```fsharp
let handleKey = function
  | "Set-Cookie" ->
    rhs ...
  | _ ->
    [key, String.con...
```

Asking because I'm curious

haf:
Looks good

src/Suave/Owin.fs
TheAngryByrd: Will do.

src/Suave/Owin.fs
TheAngryByrd: Will do

src/Suave/Owin.fs
TheAngryByrd: Yeah that makes sense.

TheAngryByrd: Should there be a const in Cookies.fs for "Set-Cookie" since it's being used around a lot?

haf: Sure, that's a good idea to save a lot of allocations... In general hoisting such strings to statics to save allocations is probably a good idea.

TheAngryByrd: Hmm, I guess you could argue it's a Header const.  Any preference to where Headers.fs/Cookies.fs?

ademar:
Thank you very much!

Issue #438, opened 2016-05-17
Suave.Html: Remove `divAttr` function family?


theprash:
Suave.Html currently has function pairs like:

```fsharp
let divAttr = tag "div"
let div = divAttr [ ]
```

I think it would be simpler to have one function:

```fsharp
let div = tag "div"
```

The first approach seems to come from WebSharper, and the latter is more like [Elm's approach](http://package.elm-lang.org/packages/elm-lang/html/1.0.0/Html).

There's a trade-off here but I think I'd prefer to type/read `[]` when there are no attributes than `Attr` when there are. Always using the same function name, and having it always match the HTML tag name, also makes things more consistent, and maybe a bit more readable for those used to HTML.

haf:
@theprash Could you PR the change and we'll review? Modifying examples to match is a plus. :)

ademar:
I think we initially had this the way you propose and someone, not sure if it was Don Syme himself,  suggested including the Attr version.

Perhaps it is safer to keep these functions around and just mark them as obsolete; there is already a lot of places out there using this notation like the Suave Music Store tutorial.

theprash:
This would involve changing the `div` functions so it would be a breaking change no matter what.

I think a predictable number of arguments is especially important for an internal DSL in a language with currrying, as you have less chance of forgetting an argument.

I'll make a PR when I get a chance.

theprash:
PR #516 submitted. I didn't find any usages of `Suave.Html` in the examples in this repo.

PR #516, opened 2016-10-08
Remove `divAttr` function family

theprash:
For #438.

I couldn't run all tests because I was getting libuv errors:

```
razor test/simple test: Exception: System.DllNotFoundException: Unable to load DLL 'libuv.dll': The specified module could not be found.
```


ademar:
Thanks!

PR #517, opened 2016-10-13
Upgrade logging facade, coloured console output

adamchester:
This PR updates to the latest [Logary.Facade](https://github.com/logary/logary/#using-logary-in-a-library), which now uses a structured logging [coloured console theme](https://github.com/logary/logary/#console-logging) by default.

This screenshot shows what the [Example](https://github.com/SuaveIO/suave/blob/8d26c61fd00431f960b627c1be0a4e5e0badf6a4/examples/Example/Program.fs#L198) console output looks like after this change:

image

Notice the structured values are highlighted (based on `Type`), making it easier to scan the output and spot the key values of interest.


ademar:
Thanks!

PR #520, opened 2016-11-01
resolves #519 - fixing locking issue with fast read and writes in web…

### LINKED ISSUE #519: Websocket chokes when sending large number of messages
Upon connection to a websocket if the websocket server sends the client 1000's of messages at some random point just stops sending. From a usage perspective all I can see if that the ws.send async inside the socket ce never returns at all and blocks the rest of the application. Tried so far two different clients and am experiencing the same issue; trying another websocket framework allows it to work as per normal.
- Client connects to a websocket endpoint
- The function provided to "handshake" webpart in the "socket" computation expressions simply continues to spam the client with small messages (e.g 1, 2, 3, 4) or a repeated message
- Eventually the client will stop receiving messages and the server will be stuck.

Adding the below test case to WebSocket.fs in Suave.Tests reproduces the issue:

```
testCase "send 10000 messages to client" <| fun _ ->
  let ctx = runWithConfig webPart

  withContext (fun _ ->
    let mre = new ManualResetEvent(false)
    let amountOfMessages = 10000
    let message = "BinaryRequest16bit"
    let echo : byte array ref = ref [||]
    let count = ref 0

    use clientWebSocket = new WebSocketSharp.WebSocket(sprintf "ws://%s:%i/websocket" ip port)

    clientWebSocket.OnMessage.Add(fun e ->
      echo:= e.RawData
      count := !count + 1
      printfn "Count is %i" !count
      if !count = amountOfMessages then mre.Set() |> ignore
    )
    clientWebSocket.Connect()
    for i = 1 to amountOfMessages do
        clientWebSocket.Send(message)

    mre.WaitOne() |> ignore
    clientWebSocket.Close()
    Assert.Equal("should be equal", testByteArray (int PayloadSize.Bit16) echo.Value , true)
    Assert.Equal("received message count on websocket", amountOfMessages, !count)
    ) ctx
```

Also created a repo to reproduce the issue - simply run the console program: https://github.com/mvkra/WebsocketFrameworkTester

Bug happens on both Windows 7 Net 4.6 and Mono 4.6.1.5 so it isn't a runtime issue. At rare times it passes and different test runs get through either more or less messages before it stalls. If you can't reproduce comment out the (printfn "Count is %i" !count) and/or up the message count required.

If any other info is required happy to assist.

mvkra:
This fixes the issue where bulk sending and or receiving can cause an inconsistent state between the queue of Async's to notify (via callbacks) and whether the read or write loop threads should continue spinning.

I've moved away from Async.WithContinuations as well replacing it with a Semaphore. This way errors will propogate up to the person sending and/or receiving easier. Performance was tested between semaphore and non semaphore (i.e locking the queue and reading/writing variables together) and saw slight improvement (if at all) with this implementation.

### DIFF
--- a/src/Suave.Tests/WebSocket.fs
+++ b/src/Suave.Tests/WebSocket.fs
@@ -156,4 +156,33 @@ let websocketTests cfg =
         clientWebSocket.Close()
         Assert.Equal("should be equal", testByteArray (int PayloadSize.Bit32) echo.Value , true)
         ) ctx
+    testCase "echo large number of messages to client" <| fun _ ->
+      let ctx = runWithConfig webPart
+
+      withContext (fun _ ->
+        let mre = new ManualResetEvent(false)
+        let amountOfMessages = 1000
+        let echo = ref []
+        let count = ref 0
+
+        use clientWebSocket = new WebSocketSharp.WebSocket(sprintf "ws://%s:%i/websocket" ip port)
+
+        clientWebSocket.OnMessage.Add(fun e ->
+          echo:= e.Data :: !echo
+          count := !count + 1
+          if !count = amountOfMessages then mre.Set() |> ignore
+        )
+        clientWebSocket.Connect()
+
+        for i = 1 to amountOfMessages do
+          clientWebSocket.Send(i.ToString())
+
+        mre.WaitOne() |> ignore
+        clientWebSocket.Close()
+
+        let expectedMessages = [ for i in amountOfMessages .. -1 .. 1 -> i.ToString() ] 
+
+        Assert.Equal("should be equal", (expectedMessages = echo.Value), true)
+        Assert.Equal("received message count on websocket", amountOfMessages, !count)
+        ) ctx
   ]
--- a/src/Suave/WebSocket.fs
+++ b/src/Suave/WebSocket.fs
@@ -13,6 +13,7 @@ module WebSocket =
   open System.Collections.Concurrent
   open System.Security.Cryptography
   open System.Text
+  open System.Threading
 
   let magicGUID = "258EAFA5-E914-47DA-95CA-C5AB0DC85B11"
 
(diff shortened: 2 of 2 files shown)


ademar:
Humm .. sorry but the fix is incorrect. You should just fix the locking issue and not change the implementation. 

You need to support multiple readers and writers which your implementation does not (remember we fixed this sometime ago on #519 ).

mvkra:
I'm not sure what you mean. 

The implementation still supports multiple readers/writers on the same "webpart" so I'm not sure what you are asking for - the semaphores are local to the websocket connection from what I can see. This is because you get a Websocket class for every websocket connection to a client. Sockets technically for each direction can't be written to by multiple writers or readers since that's what the queues were for anyway as you mentioned in the issue comments. What you are doing is exactly the same - I'm just avoiding the creation of two different threads.

The current implementation has other issues (other than the locking) mainly that errors downstream on the async's do not propogate through the thread loop construct. Any errors thrown on the async will not propogate outside the library with the current design from what I can see.

ademar:
Ah I see now. Yeah definitely this is the way to do it.

Do you want to send another PR for the 1.x branch ?

mvkra:
I tried taking this commit and putting it in the releases/v1.1x branch (this is the right one hopefully) but Websocket.fs is behind on that branch significantly. A previous websocket change (probably mine) around using the ArraySegments has no been incorporated probably because it breaks the API and warrants a major version? Will have to re purpose this fix (code may be slightly different) for the 1.1x release I'm guessing.

haf:
I suggest that we release v2 as it is right now actually...

@mvkra Can you use a prerelease nuget until it's completed?

mvkra:
I can wait certainty. We've just got our own custom "handshake" operator which we will replace with "official Suave" one once this fix is upstream. Better than having to fork the whole library in any case.

ademar:
@haf I agree, lets release v2 without HTTP2. HTTP2 support would have been "degraded" anyways because of lack of ALPN negotiation in SslStream.

haf:
I'll start by drafting release notes tomorrow then.

haf:
Put up a draft: https://github.com/SuaveIO/suave/releases/tag/untagged-515edbb75531e945ceb4

Let's solve these issues before the release as well https://github.com/SuaveIO/suave/issues?q=is%3Aopen+is%3Aissue+milestone%3Av2

The ones assigned to me are API thingies that I have laying around our code-bases. I'll fix those tomorrow. 

How about meeting on Skype tomorrow when you wake up?

PR #521, opened 2016-11-01
Place spaces between attributes

jimfoye:
The current code does not output spaces between attribute pairs. For example, it may output

<input type="submit"value="Submit"/>

Apparently browsers don't have a problem with this. I was surprised to learn that, but I assume that this is legal. However, it did cause a problem for me when I wrote some code to pretty-print Suave's output. I was trying to parse the result of xmlToString() with XDocument.Parse(), and that throws an exception when it encounters the two attribute pairs with no space between them.

In general, it would seem like a good idea to put spaces between attribute pairs, just for readability.

### DIFF
--- a/src/Experimental/Xml.fs
+++ b/src/Experimental/Xml.fs
@@ -90,29 +90,29 @@ let samplePage =
 
 /// Rendering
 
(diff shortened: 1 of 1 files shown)


ademar:
Thanks!

PR #525, opened 2016-11-04
Use expecto for all tests, refactor Testing library

haf:
This PR also includes bumping all dependencies and doing some refactoring of the tests to be more legible.

 - Use Expecto instead of Fuchu
 - Force utf-8 encoding in all files
 - Don't force newline at the end of files
 - Force tests to run sequenced
 - Bump all libraries to require .Net 4.6.1 in preparation for v2, including bumping paket monikers, /cc @forki @enricosada (see bottom of diff)
 - Bump netstandard level to 1.6 in preparation for "preview3" from MSFT
 - Remove Nancy.CSharp, because it's just a pain to get compiling on dotnet core and my time is limited. I would love to bring it back in if someone can maintain it.
 - Currently removed Perf.fs tests because we've switched from PerfUtil to BenchmarkDotNet. Now, all we have to do is follow the [docs for it](https://github.com/haf/expecto#benchmarkdotnet-usage) to start performance optimising Suave
   * Also see [this sample](https://github.com/haf/expecto/blob/master/Expecto.Tests/BenchmarkDotNet.fs)
   * [Lively discussion possible](https://gitter.im/PerfDotNet/BenchmarkDotNet)
 - Freeze Freya to beta 300 due to API breakage - /cc @panesofglass @kolektiv 

Tests are passing, but .Net core breaks the build because the Core team has moved the (downloadable) cheese again, it would seem.

/cc @ademar


haf:
Merged.

PR #529, opened 2016-11-05
Allow visual studio to build after recent project.json additions

### LINKED ISSUE #528: Suave solution doesn't work in Visual Studio
The reason is because of the new `project.json` stuff.

Output from build:
```
Restoring NuGet packages...
To prevent NuGet from restoring packages during build, open the Visual Studio Options dialog, click on the Package Manager node and uncheck 'Allow NuGet to download missing packages during build.'
Project Suave is not compatible with net45 (.NETFramework,Version=v4.5). Project Suave supports: netstandard1.6 (.NETStandard,Version=v1.6)
One or more projects are incompatible with .NETFramework,Version=v4.5.
Project Suave is not compatible with net45 (.NETFramework,Version=v4.5) / win. Project Suave supports: netstandard1.6 (.NETStandard,Version=v1.6)
One or more projects are incompatible with .NETFramework,Version=v4.5 (win).
NuGet package restore failed.
```

I'll try to fix now and submit a PR.

adamchester:
This fixes #528

### DIFF
--- a/src/Suave/project.json
+++ b/src/Suave/project.json
@@ -76,7 +76,8 @@
     },
 
     "frameworks": {
-        "netstandard1.6": {
+        "net45": {},
+		"netstandard1.6": {
             "buildOptions": {
                 "define": [
                     "DNXCORE50",
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ project.lock.json
 dotnet-dev-*
 tools/coreclr/
 dotnet_cli_install.ps1
+.vs/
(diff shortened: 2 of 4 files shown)


src/Suave/project.json
```diff
@@ -76,7 +76,8 @@
     },
 
     "frameworks": {
-        "netstandard1.6": {
+        "net45": {},
```
haf: Shouldn't this be .Net 4.6.1 as per the fsproj files?

adamchester: Apparently not
image

haf:
Thanks @adamchester !

Issue #532, opened 2016-11-14
OWIN - Do not raise on missing ClaimsPrincipal


danieljsummers:
When an OWIN application does not provide a `ClaimsPrincipal`, the [current code](https://github.com/SuaveIO/suave/blob/master/src/Suave/Owin.fs#L363) is raising a `KeyNotFoundException`.  (There is a TODO there for Core support.)  The target of this environment is the owin.RequestUser key name, which is not required per [the OWIN spec](http://owin.org/spec/spec/owin-1.0.0.html#_3.2._Environment).

The net result of this is that if the hooked OWIN application does not provide a user, Suave cannot continue.  This makes displaying login pages, where we could get a user, difficult.  :)

As it's an optional field, I believe this should return null (or None, or an empty `ClaimsPrincipal ()` object* - whatever a lens-to-nothing is supposed to return) rather than raising.  I'm happy to provide a PR to make this minor tweak, if you'll let me know what the proper default value should be. (* An empty `ClaimsPrincipal` returns `false` for `IsAuthenticated`, as I recently learned after a few hours of "research coding"...)

A simple Program.fs to demonstrate the problem:

```
open Nancy
open Nancy.Owin
open Suave
open Suave.Owin

type HomeModule () as this =
  inherit NancyModule ()
  do
    this.Get ("/", fun _ -> "Hello World" :> obj)

[<EntryPoint>]
let main argv = 
  let app = OwinApp.ofMidFunc "/" (NancyMiddleware.UseNancy (NancyOptions ()))
  startWebServer defaultConfig app
  0
```

... and the project.json...

```
{
  "version": "1.0.0-*",
  "buildOptions": {
    "debugType": "portable",
    "emitEntryPoint": true,
    "compilerName": "fsc",
    "compile": {
      "includeFiles": [
        "Program.fs"
      ]
    }
  },
  "dependencies": {
    "Nancy": "2.0.0-barneyrubble",
    "Suave": "2.0.0-rc1"
  },
  "tools": {
    "dotnet-compile-fsc":"1.0.0-preview2-*"
  },
  "frameworks": {
    "netcoreapp1.0": {
      "dependencies": {
        "Microsoft.NETCore.App": {
          "type": "platform",
          "version": "1.0.1"
        },
        "Microsoft.FSharp.Core.netcore": "1.0.0-alpha-161103"
      }
    }
  }
}
```

haf:
@danieljsummers Sure, that's a great tweak. Go ahead and PR :)

PR #533, opened 2016-11-15
Don't raise for missing ClaimsPrincipal (#532)

danieljsummers:
Change from raising an exception if the ClaimsPrincipal isn't found to simply return null


ademar:
Thanks!

Issue #539, opened 2016-11-30
Authentication.HttpContext.sessionId casts byte[] to string


radekm:
I'm using Suave 2.0.0-rc2.

`Authentication.HttpContext.sessionId` casts `userState` value which has type `byte[]` to `string`. The invalid cast is on the last line:

```fsharp
  let sessionId x =
    x.userState
    |> Map.tryFind StateStoreType
    |> Option.map (fun x -> x :?> string |> parseData)
```

It can be fixed by replacing the last line with the following:

```fsharp
Option.map (fun x -> x :?> byte[] |> UTF8.toString |> parseData)
```

Another way to fix this is to somehow change `Cookie.cookieState` that it can put values of different types (not only `byte[]`) to `userState`.

PR #540, opened 2016-11-30
Fix reading of session id

radekm:
Fixes https://github.com/SuaveIO/suave/issues/539.


radekm:
How the new test fails without the fix can be seen at https://travis-ci.org/radekm/suave/builds/180165472

ademar:
Thanks!

Issue #542, opened 2016-12-04
Incorrect handling of absolute request URIs


jbrestan:
I've encountered an HTTP proxy/API gateway which sends absolute URI on the HTTP request first line. HttpBinding.uri function, which composes the full HttpRequest.uri, handles this incorrectly by gluing the whole string to the schema, host and port again, resulting in invalid URI like "http://0.0.0.0:8080/http://example.com/path/to/resource".

I was surprised the absolute URI in HTTP requests is actually valid according to RFC 2616, so I think Suave should be able to handle this case.

jbrestan:
Added a PR to fix this case, although I'm not sure about a few things:
1) Should the resulting URI be the original absolute URI, or as it is implemented now, the binding's base URI + path from the original?
2) Can the unescaping break anything/create a security hole? Keeping things escaped broke some other tests
3) Most importantly as this is my first PR here - is this an OK way to propose a fix/extensions? :)

haf:
1. In the app, it should be the original absolute URI because the binding is for selecting the interface, not for the logical abstraction of HTTP. But for a particular method on `HttpBinding` we want the behaviour you describe.
 2. Yes, if escaping isn't done properly we can have a security hole. The app should always read escaped data unless it queries for an unescaped URI.
 3. It's a great way to start 👍

jbrestan:
PR is merged, closing. Thanks!

PR #543, opened 2016-12-04
Fix for #542

jbrestan:
Extend HttpBinding.uri to handle absolute URIs correctly as requested by https://github.com/SuaveIO/suave/issues/542

### DIFF
--- a/src/Suave.Tests/Types.fs
+++ b/src/Suave.Tests/Types.fs
@@ -62,6 +62,21 @@ let httpBinding (_ : SuaveConfig) =
       let binding = HttpBinding.create HTTP IPAddress.Any 80us
       let actual = binding.uri "http://example.com/path/to/resource" "" |> sprintf "%O"
       Expect.equal actual "http://0.0.0.0/path/to/resource" "absolute uri"
+
+    testCase "uri does not get url-encoded" <| fun _ ->
+      let binding = HttpBinding.create HTTP IPAddress.Any 80us
+      let actual = binding.uri "http://example.com/å b/c" "ö=2" |> string
+      Expect.equal actual "http://0.0.0.0/å b/c?ö=2" "uri"
+
+    testCase "url-encoded uri gets decoded" <| fun _ ->
+      let binding = HttpBinding.create HTTP IPAddress.Any 80us
+      let actual = binding.uri "a%20b" "" |> string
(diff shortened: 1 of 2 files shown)


src/Suave/Http.fs
```diff
@@ -415,7 +415,11 @@ module Http =
       socketBinding : SocketBinding }
 
     member x.uri (path : string) query =
-      let path' = if path.StartsWith "/" then path else "/" + path
+      let path' =
+        match Uri.TryCreate(path, UriKind.Absolute) with
+        | true, uri -> uri.AbsolutePath |> Uri.UnescapeDataString
```
ademar: It looks great, thanks. If you could just avoid using the pipe here we can merge it.

src/Suave.Tests/Types.fs
```diff
@@ -57,6 +57,11 @@ let httpBinding (_ : SuaveConfig) =
       let binding = HttpBinding.create HTTP IPAddress.Loopback 80us
       let actual = binding.uri "" "" |> sprintf "%O"
       Expect.equal actual "http://127.0.0.1/" "uri"
+
```
haf: I'd like another test that properly validates the whether the URI is properly escaped. As far as I can see, this uri doesn't need escaping.

jbrestan: Huh, turns out in this case the unescaping in HttpBinding.uri is unnecessary as well. I encountered a test which expected unescaped results but seems it was when I handled the incoming URI differently. Would the test suite be OK like this if I remove the unescaping from the implementation?

jbrestan: The System.Net.Uri object handles that. After parsing the URI string it will provide both escaped and unescaped data. I can write a couple of tests to check for double encoding/decoding issues and make sure the expectations are explicit.

haf: Sounds like a plan!

jbrestan: Wow, the current escaping behavior looks kind of wild, but I've added tests to describe it and to make sure my change didn't affect it. Changing it may break backwards compatibility and extend the scope of this PR so I'd rather avoid it for now.

haf:
We're a wild bunch, you know! Thanks for contributing. Let's close the PR here – we're looking forward to more fine contributions from yourself.

Issue #547, opened 2016-12-21
bug in owin handling means signalr doesn't connect


robertpi:
I'm been trying to get signalr working with suave. It seems most things go okay, but due to a bug when the connect url is called, it's treated as a 404.

There's already a fix available here: https://github.com/TheAngryByrd/suave/tree/signalrdemo

but it doesn't seem to have been merged into the master branch. Should I add the fix to my own clone of suave and send a PR? Or can a PR be send from TheAngryByrd's repo?

Thanks,
Rob

haf-lynx:
We could start by pinging @theangrybyrd to see if he has any comments on his branch and code - then if you the PR we'd be very pleased. =)

ademar:
It think what he does there is equivalent to the following, so you may don't need the patch:

```fsharp
context(fun ctxt -> owin ctx.request.path)
```

TheAngryByrd:
@ademar 
So like this? 
```
let owin path = builder.Build() |> OwinApp.ofAppFunc path
        let app = 
            
            choose [
                GET >=> path "/" >=> Files.browseFileHome "index.html"
                context(fun ctx -> owin ctx.request.path)
            ]
```

Still doesn't seem to work =\

ademar:
Hum .. yeah what I said above does not work. But I still think there is something (even if it works) incorrect with your fix because `requestPathBase`  is not the same as `ctx.request.path` at least conceptually.

ademar:
In your example `requestPathBase` should be `"/"`

TheAngryByrd:
Yeah that's what's really weird. 

If I set requestPathBase to "" it works, but if I set it to "/" it doesn't. =\

ademar:
> If I set requestPathBase to "" it works, but if I set it to "/" it doesn't. =\

We are onto something here.

TheAngryByrd:
That's with the "patch" I made to `Owin.fs`.  

If I undo the patch, the error I get from signalr is

```
 http://127.0.0.1:8083/connect?transport=longPolling&clientProtocol=1.5&conn…2F77z7EqRgPoxMuavTHf&connectionData=%5B%7B%22name%22%3A%22demohub%22%7D%5D 
```

I remember stepping through SignalR and they do some magics to come up with urls.

ademar:
> OWIN spec requires request path base to be either the empty string or start with "/"

So I guess it is not wrong to set it to `""`

ademar:
[...]

PR #548, opened 2016-12-21
Fix for #547 (Signalr)

ademar:
fix for #547


TheAngryByrd:
Since we have to supply `"/signalr"` here:
```
let owin = builder.Build() |> OwinApp.ofAppFunc "/signalr"
```

does that mean we can't chain things together on the same builder?  Like `builder.UseCookieAuthentication().MapSignalR().UseNancy()` ?

ademar:
It depends, in this case SignalR is using `requestPathBase` to derive the connect URL. See the following links:

https://github.com/aspnet/SignalR-Server/blob/9680e9c2c185a19604390003a6f10489d0538a56/src/Microsoft.AspNetCore.SignalR.Server/Hosting/RequestExtensions.cs#L10

https://github.com/aspnet/SignalR-Server/blob/9680e9c2c185a19604390003a6f10489d0538a56/src/Microsoft.AspNetCore.SignalR.Server/Transports/TransportDisconnectBase.cs#L218

So I guess in this case there is no way around it.

haf:
```
[17:44:53 ERR] e2e/Setting a path in Suave should mount an OWIN app at that path and set requestPathBase failed in 00:00:00.0137786. 

Http Status Code. Actual value was OK but had expected it to be NotFound.
```

ademar:
yeah i saw that .. I'm trying to go to the end of it; so far it seems like SignalR does not like something like:

```
owin.RequestPathBase = "/signalr" and 

owin.RequestPath = "/js"  
```

which is correct according to spec

ademar:
relevant: http://stackoverflow.com/questions/25648722/how-are-owincontext-request-path-and-pathbase-populated

ademar:
OK, this should do it.

I found that the SignalR middleware was (trying to) alter the properties `RequestPathBase` and `RequestPath` which our previous implementation with the `constant` lense was not allowing. I introduced a new `writeable` lense to make those properties read/write.

@TheAngryByrd note that now we are calling the middleware with an empty string.

```fsharp
let owin = builder.Build() |> OwinApp.ofAppFunc ""
```

TheAngryByrd:
🎸

PR #553, opened 2017-01-01
Update void elements in Suave.Html

nsmarkop:
According to the HTML5 spec, "img" and "input" are void elements:

https://www.w3.org/TR/html5/syntax.html#void-elements

They are currently defined using "tag" with empty content but using "voidTag" seems more appropriate.


ademar:
Thank you!

PR #555, opened 2017-01-03
Expand DotLiquid example.

pittsw:
This change makes it clear that DotLiquid is given an object called "model," rather than a separate object for each of the fields in Model.


haf:
Thanks!

PR #567, opened 2017-01-19
Fix a few endianness bugs on HTTP/2 branch

rmunn:
* get31bit now works on both little-endian and big-endian architectures
* get31bit now explicitly expects its byte[] arg in network byte order
* stream identifiers now encode correctly
* poke16 function now writes bytes in network byte order (big-endian)
* poke32 function now writes bytes in network byte order (big-endian)

Also stripped a few instances of trailing whitespace, in keeping with the project's whitespace policy in its .editorconfig file.

**Caution:** I have not yet been able to test these changes on a big-endian architecture; I thought that the Raspberry Pi, which uses an ARM chip, would be big-endian, but after downloading and setting up a Raspberry Pi emulator, I discovered that the Pi's chip is bi-endian (able to run in either mode), and is set to little-endian by default. So these changes are untested, but I believe them to be correct.

Certainly the `get31bit` function should be using `&&& 127`, not `&&& 128`, if the goal is to strip the most significant bit and leave the other bits alone. (And it's better to guarantee which byte it's going to be stripping, hence my change to when the `checkEndianness` call happens). The changes I'm least sure of are the changes to `poke16` and `poke32`. I believe, from what I've read about big-endian architectures, that the `byte (n >>> 24)` logic will produce the most significant byte on both little- and big-endian architectures, but that's the part I've been unable to test since I haven't yet found a big-endian system to emulate. However, I do believe that it's best if those two functions always poke their bytes into arrays in network byte order, and that's how I wrote them in this patch.

### DIFF
--- a/src/Suave/Http2.fs
+++ b/src/Suave/Http2.fs
@@ -4,7 +4,7 @@ module Http2 =
 
   let connectionPreface = "PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n"
 
-  type ErrorCode = 
+  type ErrorCode =
     | NoError
     | ProtocolError
     | InternalError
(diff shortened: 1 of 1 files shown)


ademar:
Thanks!

PR #568, opened 2017-01-19
Adding bindEmptyForm to bind forms without any predefined validations.

tamizhvendan:
A syntactic sugar function to do model binding that doesn't need to use the predefined validations. I found it useful in cases where we rely on domain level validations.

### DIFF
--- a/src/Experimental/Form.fs
+++ b/src/Experimental/Form.fs
@@ -144,7 +144,7 @@ let bindForm<'a> (form : Form<'a>) (req : HttpRequest) =
     return record
   }
 
-let bindForm2<'a> = 
+let bindEmptyForm<'a> = 
   let form : Form<'a> = Form([],[])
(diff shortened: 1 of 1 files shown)


haf:
Can we replace the `2` for something more semantically meaningful?

tamizhvendan:
@haf I've felt the same but couldn't able to come out with a better name so went with the F# standard library [way](https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/list.map2%5B't1,'t2,'u%5D-function-%5Bfsharp%5D). How about `bindFormWithOutValidation` or `bindJustForm`?

ademar:
`bindEmptyForm` ?

haf:
`bindEmptyForm` +1

tamizhvendan:
Renamed to `bindEmptyForm`

PR #575, opened 2017-02-07
Fix Access-Control-Allow-Credentials casing

vbfox:
According to [the specification](https://www.w3.org/TR/cors/#http-access-control-allow-credentials) this header can only have one valid value: 'true' (Lower case).

>     Access-Control-Allow-Credentials: "Access-Control-Allow-Credentials" ":" true
>     true: %x74.72.75.65 ; "true", case-sensitive

The correct value is now emitted when allowCookies is set and the header omitted otherwise.

Chrome is now checking this header and consider the current values emitted by Suave as invalid (And so refuse all XHR connections)


ademar:
Thanks!

PR #577, opened 2017-02-07
Update dotliquid.md

jimfoye:
Enhanced docs for using Suave.DotLiquid.


ademar:
Thanks you!

jimfoye:
FYI the page doesn't look right on the website. I think the docs themselves must at some point be processed by DotLiquid, as it's all the DotLiquid syntax that is getting stripped out.

PR #578, opened 2017-02-10
Allow further customising the console output

adamchester:
This bumps the Logary.Facade version to the latest. See logary/logary#234 for changes.

Using the new `LiterateConsoleTarget` ctor, with the `outputTemplate` parameter, one can configure the console output in a way that is very similar to what older versions of Suave supported:
screen shot 2017-02-10 at 12 09 24

Notably, until now, it was not possible to render the log event `[{source}]`.


haf:
Thank you @adamchester!

PR #580, opened 2017-02-17
Basic authentication: support asynchronous authentication check

the-g-man:
Basic authentication would commonly require IO - e.g. to look up a password hash in a database by username (this is my motivating use case). The `authenticateBasic` function only provides the ability to authenticate a (username, password) tuple by a simple blocking call. This PR adds an `authenticateBasicAsync` function which takes a non-blocking password check function instead.


ademar:
Looks great. Thank you very much!

PR #584, opened 2017-03-01
fix clean checkout build to use Logary.Facade instead of Suave.Logging

mvkra:
Currently when you checkout Suave the build fails due to the change in namespace of Facade.fs.

Considering this is upgraded to be more generic (Logary.Facade) vs (Suave.Logging) - I'm guessing the intention was to port Suave to a more generic facade that can be used by multiple projects. So I've changed Suave's logging to the new namespace in this PR.

Afterwards the repo builds on a clean checkout which is important for doing any other open source PRs.

### DIFF
--- a/examples/Asp.NET/Program.fs
+++ b/examples/Asp.NET/Program.fs
@@ -1,6 +1,6 @@
 ﻿
 open Suave
-open Suave.Logging
+open Logary.Facade
 open System.IO
 open System.Net
 
--- a/examples/CORS/Program.fs
+++ b/examples/CORS/Program.fs
@@ -4,7 +4,7 @@ open System
 open System.Net
 
 open Suave
-open Suave.Logging
+open Logary.Facade
 open Suave.Filters
 open Suave.Writers
 open Suave.Files
--- a/examples/Example/Program.fs
+++ b/examples/Example/Program.fs
@@ -5,7 +5,7 @@ open System.Net
 
 open Suave
 open Suave.Sockets.Control
-open Suave.Logging
+open Logary.Facade
 open Suave.Operators
 open Suave.EventSource
 open Suave.Filters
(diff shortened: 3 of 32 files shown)


ademar:
Hum .. as far as I know the build is not broken, are you running the Rake script ?

mvkra:
Yep check checkout then running rake and or build.sh. not sure why the
build server is working. Maybe he facade.fs is an old one? It's broken on
two different machines for me so far.

adamchester:
@mvkra to build you must use `bundle install && bundle exec rake`. This [line in the Rakefile](https://github.com/SuaveIO/suave/blob/8c2ff7aa5ab93233521f7e211fc702bb787d6e96/Rakefile#L26) changes the namespace.

Also note that even we wanted, we cannot change the namespace from `Suave.Logging` to `Logary.Facade` without a major version change - because it breaks all users.

mvkra:
Ok the "bundle install" I think is what was missing in the build instruments. i.e this is what's on the main page. If I do this it fails.  Lack of Ruby knowledge I guess biting me here.

What's on the main page right now:

To execute the build script, invoke following commands:

bundle // "should this be "bundle install""?
bundle exec rake

adamchester:
I think the bug is with [`build.sh`](https://github.com/SuaveIO/suave/blob/8c2ff7aa5ab93233521f7e211fc702bb787d6e96/build.sh#L26) - it does this:

```bash
echo 'Restoring nugets'
run tools/paket.exe restore

echo 'Building'
xbuild src/Suave.sln /p:Configuration=Release

echo 'Running tests'
mono src/Suave.Tests/bin/Release/Suave.Tests.exe --sequenced
```

If `paket restore` updates the `Facade.fs` then it brings in one with the `Logary.Facade` namespace.

mvkra:
I'm closing this PR. It was caused because I didnt run "bundle install" and was using older or missing dependencies that did the namespace change I believe.

Out of this though I think the "How to Build" doco needs to be updated to use "bundle install" rather than just "bundle".

PR #586, opened 2017-03-01
allow reuse of byte buffer when reading from a websocket

mvkra:
As part of https://github.com/SuaveIO/suave/pull/485 in order to decrease large object heap fragmentation I added our code to reuse byte buffers for sending to avoid memory pressure we experienced on our servers. However I didn't get around to putting up a PR for the reading side meaning the API for reading/writing currently doesn't match well.

This PR has two points: 

- Allows you to provide your own byte segment for reading via a function. There was a number of ways to do this and I'm open to feedback here on other methods (e.g allow use to provide a stream instead, using the BufferManager and exploiting the fin flag to get the user to keep reading, method returns a stream the user disposes, etc). The heart of the issue is that you don't know the length in advance before reading. 
- Avoid another allocation arrays when applying the mask.

I've added another method since I'm assuming we are respecting SemVer here and probably don't want to break the API now that v2 is released. I haven't changed array allocations for the headers and such since these arrays are typically small and won't make it to the LOH.

### DIFF
--- a/src/Suave.Tests/WebSocket.fs
+++ b/src/Suave.Tests/WebSocket.fs
@@ -32,6 +32,9 @@ let websocketTests cfg =
 
   let runWithConfig = runWith cfg
 
+  let reuseableBuffer = Array.zeroCreate 1000
+
+  let websocketAppUrl = "/websocket"
   let websocketApp (webSocket : WebSocket) =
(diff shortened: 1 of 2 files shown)


ademar:
Thanks!

PR #590, opened 2017-03-16
use .net core sdk 1.0 rtm

enricosada:
the `Microsoft.FSharp.Core.netcore` is deprecated to use directly.
use the [FSharp.Core](https://www.nuget.org/packages/FSharp.Core/) package, v4.1 now support netstandard1.6


enricosada:
let me test this a bit more 😄 , because a new drop of fsharp.core is near, just to be sure

haf:
Of course :)

enricosada:
@haf i'll just convert everything to sdk 1.0 rtm and new fsproj

enricosada:
@ademar @haf i think is ok.

- i migrate to .net core sdk 1.0 (so fsproj instead of project.json).
- use `FSharp,Core` instead of `MS.Fsharp.Core.netcore` (deprecated)
- fixed build script (with update install script)
- tested locally 
- tested it with https://github.com/SuaveIO/Suave-CoreCLR-sample too (updated)

i did not test the install script on osx/ubuntu, but now is the same as windows

ademar:
Thanks Enrico, much appreciated!

enricosada:
np @ademar, any issue feel free to ping me directly.
If is possibile to have a Suave `2.0.5-alpha` of this, will be really helpful, because we are trying to move FsAutocomplete to full .net core. 
Atm i cannot use `2.0.4` because it depends on `MS.Fsharp.Core.netcore` nupkg, so i cannot mix it with other packages using `Fsharp.Core` 4.1 nupkg, otherwise i get two FSharp.Core.dll referenced 😄 

That will enable all FSAC based tools (vscode/emac/vi) to run with just .net core installed (without .net or mono required)

ademar:
Sure thing, I'll get to it

ademar:
@enricosada just pushed 2.0.5-alpha.

Issue #597, opened 2017-03-28
Improve Content-Range by specifying length


rmunn:
The [HTTP spec](http://httpwg.org/specs/rfc7233.html#header.content-range) says:

> For byte ranges, a sender SHOULD indicate the complete length of the representation from which the range has been extracted, unless the complete length is unknown or difficult to determine. An asterisk character ("`*`") in place of the complete-length indicates that the representation length was unknown when the header field was generated.

Compression makes this tricky, so I'm okay with not specifying the length of the original file if `Compression.transformStream` returns `Some`. But if the file is being sent uncompressed, it should be straightforward to specify its complete length in the `Content-Range` header. And since that's a SHOULD in the HTTP spec, we should try to implement it if we can.

Issue #596, opened 2017-03-28
Off-by-1 error in new Content-Range code


rmunn:
I'm glad to see the new Content-Range code, but I believe there's an off-by-1 error in the headers it sends. The [`sendFile` function](https://github.com/SuaveIO/suave/blob/d997c52f19bcbf311e6b953f0de00e879771d8bf/src/Suave/Combinators.fs#L527) currently sends a `Content-Range` header specifying the range from `start` to `start + fs.Length`, so if we're sending 500 bytes starting at index 0, it would send `Content-Range: bytes 0-500/*`, and 500 bytes starting at index 500 would be `Content-Range: bytes 500-1000/*`. But [the HTTP spec for Content-Range](https://tools.ietf.org/html/rfc7233#section-4.2) specifies the end of the range as "`last-byte-pos`", and the examples show sending 500 bytes by using `Content-Range: bytes 0-499/*` or `Content-Range: bytes 500-999/*`. I believe the end byte in the Content-Range should be calculated as `start + fs.Length - 1` to comply with the HTTP spec. (Two changes needed in `sendFile` as it is currently written).

Note that I have not yet tested this with actual browsers, so I don't know if this is a serious error or if browsers would be able to work around it. This is based simply on reading the code and the HTTP spec.

ademar:
ping @simonhdickson

simonhdickson:
Ah yes, I haven't noticed any issues with browsers but it's likely still worth correcting.

PR #598, opened 2017-03-29
Minor Range content fixes

simonhdickson:
Fixes for #596 and #597 also returns 206 for partial content rather than 200 (I don't think most clients care as long as you're in the the 2XX range but it is in the spec). I don't know if it is bad to move the Stream creation outside of the Socket but I've ran it locally it seems to work the same as before.

### DIFF
--- a/src/Suave/Combinators.fs
+++ b/src/Suave/Combinators.fs
@@ -489,14 +489,13 @@ module ContentRange =
     | Choice1Of2 rangeValue -> Some <| parseContentRange rangeValue
     | Choice2Of2 _ -> None
 
-  let getFileStream (ctx:HttpContext) =
+  let getFileStream (ctx:HttpContext) path =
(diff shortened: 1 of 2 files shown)


ademar:
Gracias!

rmunn:
One possible issue I can see with moving the Stream creation outside the `socket` CE is closures. Specifically, I wonder what would happen in the following scenario:

1. Create `test.txt` with contents "Hello, world".
1. Run Suave and load `test.txt` in your browser via `sendFile`.
1. Edit `test.txt` to say "Hello, Suave!"
1. Refresh your browser. What happens?

I don't actually know that there will be a problem — I don't have the time right now to go through the implementation with a fine-toothed comb — but I could see there potentially being a subtle race-condition issue, depending on when the `socket` CE closes over the FileStream, and what happens to it in the meantime.

haf:
@simonhdickson Could you answer rmunn?

simonhdickson:
I haven't seen any issues so far but I think there might be a way to avoid creating the stream outside the socket.

PR #601, opened 2017-04-07
Fix two endianness issues in WebSocket code

rmunn:
[RFC 6455](https://tools.ietf.org/html/rfc6455) specifies that multi-byte integers must be sent in network byte order (i.e., big-endian). The length-encoding code in WebSocket.fs correctly sends bytes in network byte order on little-endian machines, but would send them in the wrong order if it were ever run on big-endian machines. This is a bit theoretical as it's currently rather hard to find a big-endian machine, but it should be fixed anyway.

The status code, on the other hand, would be correct on a big-endian machine but is encoded in the wrong byte order on a little-endian machine, so that definitely needs fixing.

The `.editorconfig` file's setting of `trim_trailing_whitespace = true` produced a lot of incidental whitespace fixes when I saved the file. I chose to leave them in the PR since you probably want them applied, but I can trim them to produce a smaller diff if that's desirable. (See, for example, my first commit where I trimmed the whitespace changes so that it would be quite obvious what I had changed and what I hadn't).

### DIFF
--- a/src/Suave/WebSocket.fs
+++ b/src/Suave/WebSocket.fs
@@ -127,10 +127,10 @@ module WebSocket =
         match data.Count with
         | l when l >= 65536 ->
           [| yield 127uy
-             yield! BitConverter.GetBytes(uint64 data.Count) |> Array.rev |]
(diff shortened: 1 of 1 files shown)


ademar:
Thanks!

PR #602, opened 2017-04-07
Html.fs: HTML-encode attribtue values & text of elements

atifaziz:
This PR encode attribute values and text of elements in `Html.fs`.

Example of invalid HTML before this PR:

```
> p ["title", "\"foo\" and \"bar\""] (text "foo & bar ") |> htmlToString;;
val it : string = "<p title=""foo" and "bar"">foo & bar </p>"
```

Correct encoding after applying this PR:

```
> p ["title", "\"foo\" and \"bar\""] (text "foo & bar ") |> htmlToString;;
val it : string = "<p title="&quot;foo&quot; and &quot;bar&quot;">foo &amp; bar </p>"
```

### DIFF
--- a/src/Experimental/Html.fs
+++ b/src/Experimental/Html.fs
@@ -1,6 +1,7 @@
-module Suave.Html
+﻿module Suave.Html
 
 open System
+open System.Net
 
 type Attribute = string * string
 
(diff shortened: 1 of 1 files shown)


ademar:
Thanks!

Issue #610, opened 2017-04-23
Html.fs: need way to avoid HtmlEncode in Text nodes


vivainio:
I observed that Text nodes now get HtmlEncode'd:

https://github.com/SuaveIO/suave/commit/9d85fb35a15df708335905c0bb7fbc466287b28a (by @atifaziz).

This is desirable behavior for most cases, but for text in <script> tags this is not wanted.

Type "Node" probably needs to grow a new discriminated union case RawText, that doesn't perform the encoding, to allow creating scripts with 

```fsharp
tag "script" [] <| rawText ("Hello world <<>>")
```

vivainio:
I made a PR for this: https://github.com/SuaveIO/suave/pull/611

haf:
Thanks, I'm rebuilding currently.

PR #611, opened 2017-04-23
Html.fs: add 'rawText' as a non-html-escaping alternative to 'text'

vivainio:
Add "rawText" to Html.fs, that passes content through without &quot; etc. HTML encoding.

Encoding broke my generated <script> tags, so it's quite important to fix it. I can change naming or whatever as you see fit.

Fixes https://github.com/SuaveIO/suave/issues/610


haf:
Thanks!

Issue #608, opened 2017-04-21
statefulForSession throws InvalidDataContractException


danieljsummers:
I've added "statefulForSession" in front of my `choose` for routes, and I'm now seeing the following error when I try to go to the home page of my application.

```
[07:43:17 WRN] TCP request processing failed
System.Runtime.Serialization.InvalidDataContractException: Type
'Microsoft.FSharp.Collections.FSharpMap`2[System.String,System.Object]' cannot be serialized. Consider
marking it with the DataContractAttribute attribute, and marking all of its members you want serialized
with the DataMemberAttribute attribute. Alternatively, you can ensure that the type is public and has a
parameterless constructor - all public members of the type will then be serialized, and no attributes will
be required.
   at System.Runtime.Serialization.DataContract.DataContractCriticalHelper.ThrowInvalidDataContractException(String message, Type type)
   at System.Runtime.Serialization.DataContract.DataContractCriticalHelper.CreateDataContract(Int32 id, RuntimeTypeHandle typeHandle, Type type)
   at System.Runtime.Serialization.DataContract.DataContractCriticalHelper.GetDataContractSkipValidation(Int32 id, RuntimeTypeHandle typeHandle, Type type)
   at System.Runtime.Serialization.DataContract.GetDataContract(RuntimeTypeHandle typeHandle, Type type, SerializationMode mode)
   at System.Runtime.Serialization.Json.DataContractJsonSerializerImpl.InternalWriteObjectContent(XmlWriterDelegator writer, Object graph)
   at System.Runtime.Serialization.Json.DataContractJsonSerializerImpl.InternalWriteObject(XmlWriterDelegator writer, Object graph)
   at System.Runtime.Serialization.XmlObjectSerializer.WriteObjectHandleExceptions(XmlWriterDelegator writer, Object graph, DataContractResolver dataContractResolver)
   at System.Runtime.Serialization.Json.DataContractJsonSerializerImpl.WriteObject(XmlDictionaryWriter writer, Object graph)
   at System.Runtime.Serialization.Json.DataContractJsonSerializerImpl.WriteObject(Stream stream, Object graph)
   at Suave.JsonFormatterSerialiser.Suave-CookieSerialiser-serialise(FSharpMap`2 m)
   at Suave.State.CookieStateStore.stateful@82-2.Invoke(Unit unitVar0)
   at Suave.Cookie.cookieState@236.Invoke(HttpContext ctx)
[...]

PR #613, opened 2017-04-26
Session and State Docs

### LINKED ISSUE #482: Document state-passing between web parts
I don't like to use a mutable variable here https://github.com/vasily-kirichenko/FsAutoComplete/blob/refactoring/src/FsAutoComplete.Suave/FsAutoComplete.Suave.fs#L41, which is reassigned in this code  https://github.com/vasily-kirichenko/FsAutoComplete/blob/refactoring/src/FsAutoComplete.Suave/FsAutoComplete.Suave.fs#L74

So, we have a bunch of handlers like this:

``` fsharp
POST >=>
        choose [
            path "/parse" >=> handler (fun (data : ParseRequest) -> commands.Parse state data.FileName data.Lines)
            //TODO: Add filewatcher
            path "/project" >=> handler (fun (data : ProjectRequest) -> commands.Project state data.FileName DateTime.Now false)
            path "/declarations" >=> handler (fun (data : DeclarationsRequest) -> commands.Declarations state data.FileName)
...
```

Each handler may modify (create new) `state`, then assigns it to a mutable variable. It's not very functional.

danieljsummers:
This is the document as discussed in #608 and should also close #482 .  Additionally, I synchronized `SuaveConfig` in the docs with what's in the repository, and I added "raw" Liquid tags around the DotLiquid examples; they weren't displaying correctly because Jekyll was actually interpreting them.

No code changes are in this PR; it's all docs.


docs/sessions.md
```diff
@@ -0,0 +1,127 @@
+---
+layout: default
+---
+
+State and Sessions in Suave
+===========================
+
+While stateless web applications are all the rage, there are still times where state needs to be preserved, either across requests in a session, or simply among `WebPart`s as the request/response pipeline is composed.  In these cases, Suave has you covered!  We'll look at three aspects of session cookies first, then how we can construct state among `WebPart`s.
```
haf: But even if you persist across sessions (in the cookie storage) the web app **is** stateless in the meaning intended when people speak of stateless web apps. Similarly, a pipeline as part of processing the request always has some state. I think this may confound the matter more than it helps.

docs/sessions.md
```diff
@@ -0,0 +1,127 @@
...
+
+Suave provides storage interfaces for cookies in the `Suave.State.CookieStateStore` module.  The `statefulForSession` WebPart can be composed to make a path session-aware.  From there, the `HttpContext.state` function extracts the state information, and the `get` and `set` functions on the resulting `StateStore` object can be used to manipulate the contents of the state tracked by the session.
+
+{% highlight fsharp %}
+open Suave
+open Suave.State.CookieStateStore
+
+/// string -> 'T -> WebPart
+let setSessionValue key value = context (fun ctx ->
+  match ctx |> HttpContext.state with
```
haf: Don't pipe to functions of a single argument, please.

docs/sessions.md
```diff
@@ -0,0 +1,127 @@
...
+Cookie State Session Storage
+----------------------------
+
+Suave provides storage interfaces for cookies in the `Suave.State.CookieStateStore` module.  The `statefulForSession` WebPart can be composed to make a path session-aware.  From there, the `HttpContext.state` function extracts the state information, and the `get` and `set` functions on the resulting `StateStore` object can be used to manipulate the contents of the state tracked by the session.
+
+{% highlight fsharp %}
+open Suave
+open Suave.State.CookieStateStore
+
+/// string -> 'T -> WebPart
```
haf: Add as type annotations instead?

docs/sessions.md
```diff
@@ -0,0 +1,127 @@
...
+let setSessionValue key value = context (fun ctx ->
+  match ctx |> HttpContext.state with
+  | Some state -> state.set key value
```
haf: Newlines after `->`

docs/sessions.md
```diff
@@ -0,0 +1,127 @@
...
+  | _ -> never // fail)
```
haf: You've commented out your ending `)`

docs/sessions.md
```diff
@@ -0,0 +1,127 @@
...
+
+/// HttpContext -> string -> 'T option
```
haf: Same; use the type system instead.

docs/sessions.md
```diff
@@ -0,0 +1,127 @@
...
+let getSessionValue ctx key =
+  match ctx |> HttpContext.state with
```
haf: Same

docs/sessions.md
```diff
@@ -0,0 +1,127 @@
...
+  | Some state -> state.get key
```
haf: Same.

docs/sessions.md
```diff
@@ -0,0 +1,127 @@
...
+  | _ -> None
+
+/// HttpContext -> string -> string
```
haf: Same. Maybe actually document why one would want a function that returns the empty string if the state stored-value isn't found?

docs/sessions.md
```diff
@@ -0,0 +1,127 @@
...
+let getStringSessionValue ctx key = 
+  match getSessionValue ctx key with
+  | Some value -> string value
+  | _ -> ""
+
+/// WebPart
+let cookieYes = warbler (fun ctx -> OK (getStringSessionValue ctx "test"))
```
haf: You could use `context` here for clarity and code completion support in the lamda body.

docs/sessions.md
```diff
@@ -0,0 +1,127 @@
...
+        path "/yes" >=> cookieYes
+        path "/no" >=> cookieNo
+        RequestErrors.NOT_FOUND
+        ]
+{% endhighlight %}
+
+Server Keys
+-----------
+
+The contents of the cookie are encrypted before the cookie is sent. Suave's default configuration generates a new server key each time the server is restarted. While this is not _wrong_, users would likely get quite annoyed if they lost their state because the server was restarted. Additionally, specifying a server key lets load-balanced servers access the same information.
```
haf: Well it's not wrong, that's true ;), it's to make the developer understand they need to **change** the key; otherwise **all** Suave apps would be running the same key and that would be a security vulnerability. Perhaps add that to the description?

docs/sessions.md
```diff
@@ -0,0 +1,127 @@
...
+
+Continuing our example from above... _(This should not be used on production servers; a key should be generated and provided via a configuration file.)_
```
haf: I disagree; perhaps link to https://12factor.net/ or https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables or https://www.vaultproject.io/ or ...

docs/sessions.md
```diff
@@ -0,0 +1,127 @@
...
+
+{% highlight fsharp %}
+let suaveCfg =
+  { defaultConfig with
+      serverKey = System.Text.Encoding.UTF8.GetBytes("12345678901234567890123456789012")
```
haf: Show how to generate a key in a FSX file instead. Otherwise people will copy n paste this example.

docs/sessions.md
```diff
@@ -0,0 +1,127 @@
...
+
+type JsonNetCookieSerialiser () =
+  interface CookieSerialiser with
+    member x.serialise m =
+      utf8.GetBytes (JsonConvert.SerializeObject m)
+    member x.deserialise m =
+      JsonConvert.DeserializeObject<Map<string, obj>> (utf8.GetString m)
+{% endhighlight %}
+
+Then, modify the configuration to use that serializer.
```
haf: Good

docs/sessions.md
```diff
@@ -0,0 +1,127 @@
...
+
+{% highlight fsharp %}
+/// Again, do not make this your server key in production...
+let suaveCfg =
+  { defaultConfig with
+      serverKey = System.Text.Encoding.UTF8.GetBytes("12345678901234567890123456789012")
```
haf: Same.

docs/sessions.md
```diff
@@ -0,0 +1,127 @@
...
+      cookieSerialiser = new JsonNetCookieSerialiser()
+    }
+{% endhighlight %}
+
+State among WebParts
+--------------------
+
+Within the `Writers` module, Suave provides the functions `setUserData` and `unsetUserData` for adding items to the context's `userState` property.  The example below could be used to accrue a list of messags to be displayed to the user.
```
haf: `messages` – spelling.

docs/sessions.md
```diff
@@ -0,0 +1,127 @@
...
+
+{% highlight fsharp %}
+/// string -> WebPart
```
haf: Move types to code.

docs/sessions.md
```diff
@@ -0,0 +1,127 @@
...
+let addUserMessage message =
+  context (fun ctx ->
+    let existing =
+      match ctx.userState.ContainsKey "messages" with
```
haf: I prefer if we do `match ctx.userState |> Map.tryFind "messages" with`.

docs/sessions.md
```diff
@@ -0,0 +1,127 @@
...
+      | true -> ctx.userState.["messages"] :?> string list
```
haf: Extract to `let read context key: 'value = context.userState |> Map.tryFind key |> Option.map (fun x -> x :?> 'value)` function? Then curry in example; `let read = read ctx` and use `read "messages"` which is then automatically typed and downcast to the right type.

docs/sessions.md
```diff
@@ -0,0 +1,127 @@
...
+  choose [
+    path "/"
+      >=> addUserMessage "It's a state!"
+      >=> addUserMessage "Another one"
+      >=> context (fun ctx -> Successful.OK (View.page ctx.userState))
+      >=> Writers.unsetUserData "messages"
+    ]
+{% endhighlight %}
+
+In this example, `View.page` is a function that generates the output, using the user state `Map<string, obj>` to display the messages in a nice way.  Additionally, this example shows a few other interesting aspects.  First, the user state may persist across requests; if that is not desirable for your use case, you will want to `unset` the items when they are no longer needed.  Second, you can combine `WebPart`s even after the one that sets the output content.  If you are used to an MVC environment, you can't do much after you `return` your result; these combinators let you modify the context even once the output has been generated.
```
haf: `userState` doesn't persist across requests AFAIK? So you don't have to unset anything. Also, people often think it's available if you have:

```fsharp
let app =
  choose [
    path "/a" >=> addUserMessage "It's a state!"
    path "/b" >=> context (fun ctx -> Successful.OK (View.page ctx.userState))
  ]
```

but because `userState` is not persistent, `/b` never sees the value. This could be useful to add to the docs as an example.

danieljsummers: In my testing, `userState` was persisted across requests; if I appended a message each time, the list got longer by 1 each time.  In the example, with 2 messages, I could hit "refresh" on the browser and see my messages go from 2, to 4, to 6, etc.  I'll do up a gist to demonstrate it.

I'll get the changes added to the branch shortly.

docs/sessions.md
danieljsummers: How about

While web applications are technically stateless, there are still times where we need to start from a known state.  So, let's jump in and look at aspects of session cookies first, then how we can manage state among `WebPart`s.

docs/sessions.md
ademar: > In my testing, userState was persisted across requests

I think this could be happening due to Keep-alive, the request record and the userState dictionary are being reused. I haven't looked at the code to verify it but it is possible.

docs/sessions.md
danieljsummers: Additionally, I could also just use `defaultArg` here.

docs/sessions.md
danieljsummers: Isn't the generated key random?  IOW, with the defaults, all Suave apps aren't all using the same key, just an inconsistent one.  (Just trying to make sure I understand it well.)

I'll reword it to emphasize that this is an important change to make, and capture some of the environment variable / config file stuff as well.  I've got uncommitted stuff in the same fork on the ServerKey stuff, so my next PR will tweak this section a bit.

docs/sessions.md
danieljsummers: @ademar You're exactly right; that's what's happening.  [This gist](https://gist.github.com/danieljsummers/fb2bf9427bf42191b7b9cc1ca74f2be3) is my test case.  I searched through the code for keep-alive, and and I found where it's determined in the httpLoop, but I didn't see any place to configure the timeout (or even what the timeout was).

I can mention keep-alive that paragraph; in fact, I'm glad we had this discussion because I really didn't have the cookie state vs. userData lifetimes straight in my own head.  :)

haf: Then we've discovered a bug, right?

danieljsummers: I know that was more for Ademar; I'll watch this conversation. A "gotcha" is one thing, but an "unintentional gotcha" is pretty much the definition of a bug.  :)  If it is determined that it is a bug, I'll axe that that `unsetUserData` line and fix the verbiage below it to explain the per request/per session lifetimes.

One consideration, though, is that with its current behavior, if people are using the post/redirect/get sequence of requests, `userData` is a spot to stash things that only need to persist in the context of this sequence. If that is changed, we'd want to make sure people understood that this unintended functionality isn't going to work the way it used to.

I stand ready to document your decision.  :)  _(...and the reasoning behind it)_

haf: The problem with their code if they do that is that HTTP is stateless except for cookies and each consecutive request can go to d a different node in a cluster of web servers.

danieljsummers: Ah - got it.  I'll be on the road most of the day today, but this evening (US Central) I can update it accordingly.  Would you like me to open an issue for the `userData` persistence problem, then mention it as a workaround in the docs for this PR?  Then, when that issue is worked, those notes could be removed.

haf: Yes, please. That would be much appreciated.

danieljsummers:
I've addressed all the "changes requested" - this one is ready to re-review.

docs/sessions.md
haf: Yes, it's random, but it can't be if we want the property to have "working-across-restart-cookies by default"; or else the Suave server would have to persist its newly generated key to disk, and that goes against the "just a function to run" idea. So you're right.

docs/sessions.md
```diff
@@ -56,12 +57,23 @@ Server Keys
 
 The contents of the cookie are encrypted before the cookie is sent. Suave's default configuration generates a new server key each time the server is restarted. While this is not _wrong_, users would likely get quite annoyed if they lost their state because the server was restarted. Additionally, specifying a server key lets load-balanced servers access the same information.
 
-Continuing our example from above... _(This should not be used on production servers; a key should be generated and provided via a configuration file.)_
+That being said, specifying a good server is important for security as well. You can use [a web-based service](https://www.random.org/strings/) to generate 32 characters of random text. You can also run the following, either in an .fsx file (for the full .NET Framework) or in a console app (for .NET Core), to generate a UTF-8 key. (This requires no Suave libraries.)
+
+{% highlight fsharp %} 
+let rnd = System.Random ()
```
haf: This random generator is insecure.

Why can't we call the built in `generateKey` function? https://github.com/SuaveIO/suave/blob/master/src/Suave/Utils/Crypto.fs#L61

docs/sessions.md
```diff
@@ -76,44 +88,46 @@ Cookie Serialization<br>_(of particular interest to .NET Core < netstandard2.0)_
 Suave uses the .NET Framework type `BinaryFormatter` to serialize the `Map<string, obj>` containing the session state; this is the default. However, the `BinaryFormatter` was removed in the .NET Core API, and the `DataContractJsonSerializer` does not recognize the `Map<string, obj>` type. One option is to utilize JSON.NET to serialize this object. To use that, ensure you've added the `Newtonsoft.Json` NuGet package to your project, then put the following code somewhere before the `suaveCfg` definition in the example above.
 
 {% highlight fsharp %}
-/// alias
-let utf8 = System.Text.Encoding.UTF8
-
+// NOTE: UTF8 is in the auto-opened module YoLo.
```
haf: I prefer the previous version, as YoLo is internal by default https://github.com/SuaveIO/suave/blob/master/Rakefile#L24

docs/sessions.md
```diff
@@ -76,44 +88,46 @@ Cookie Serialization<br>_(of particular interest to .NET Core < netstandard2.0)_
...
 {% endhighlight %}
 
 Then, modify the configuration to use that serializer.
 
 {% highlight fsharp %}
-/// Again, do not make this your server key in production...
 let suaveCfg =
   { defaultConfig with
-      serverKey = System.Text.Encoding.UTF8.GetBytes("12345678901234567890123456789012")
+      serverKey = UTF8.bytes "[the-generated-key]"
```
haf: When using the built-in function, there's no need to represent the bytes as chars, thus no need to convert from utf8.

docs/sessions.md
```diff
@@ -56,12 +57,23 @@ Server Keys
 
 The contents of the cookie are encrypted before the cookie is sent. Suave's default configuration generates a new server key each time the server is restarted. While this is not _wrong_, users would likely get quite annoyed if they lost their state because the server was restarted. Additionally, specifying a server key lets load-balanced servers access the same information.
 
-Continuing our example from above... _(This should not be used on production servers; a key should be generated and provided via a configuration file.)_
+That being said, specifying a good server is important for security as well. You can use [a web-based service](https://www.random.org/strings/) to generate 32 characters of random text. You can also run the following, either in an .fsx file (for the full .NET Framework) or in a console app (for .NET Core), to generate a UTF-8 key. (This requires no Suave libraries.)
+
+{% highlight fsharp %} 
+let rnd = System.Random ()
+Seq.initInfinite (fun _ -> rnd.Next (35, 126) )
```
haf: This reduces the effective key-space from 32 bytes to 11,7 bytes, weakening security by 2/3rds.

docs/sessions.md
```diff
@@ -124,4 +138,4 @@ let app =
     ]
 {% endhighlight %}
```
haf: Generated, but not yet sent to the socket.

docs/sessions.md
```diff
@@ -56,12 +57,23 @@ Server Keys
 
 The contents of the cookie are encrypted before the cookie is sent. Suave's default configuration generates a new server key each time the server is restarted. While this is not _wrong_, users would likely get quite annoyed if they lost their state because the server was restarted. Additionally, specifying a server key lets load-balanced servers access the same information.
 
-Continuing our example from above... _(This should not be used on production servers; a key should be generated and provided via a configuration file.)_
+That being said, specifying a good server is important for security as well. You can use [a web-based service](https://www.random.org/strings/) to generate 32 characters of random text. You can also run the following, either in an .fsx file (for the full .NET Framework) or in a console app (for .NET Core), to generate a UTF-8 key. (This requires no Suave libraries.)
```
haf: This needs to be updated. We want 32 random bytes. UTF8 has nothing to do with it.

docs/sessions.md
danieljsummers: I knew this, but I was trying to generate a key that would be 32 characters.  I just committed a change that uses `Crypto.generateKey` and base-64 encodes the result, storing it in a text file.  Then, they key can be moved to whatever configuration needs it.

danieljsummers:
Back to you gents...

haf:
Thank you @danieljsummers, will deploy when at laptop.

Issue #536, opened 2016-11-21
API Reference link redirects circularly


Biarity:
If you go to suave.io/Suave.html (by clicking the 'API Reference' button), then click on the API Reference button at the bottom of the page (last word of the paragraph), you will be redirected to the same page.

I didn't make a PR because I don't know what the intended page is. Is there even an API Reference?

haf:
Yes there is, but the doc gen has broken with an indecipherable error (see the Rakefile). Sorry about that...

haf:
```
mono packages/docs/FsLibTool/tools/FsLibTool.exe src docs/_site
Generating reference for Suave.
Failed to resolve: HttpContext
Failed to resolve: HttpContext
Failed to resolve: HttpContext
Failed to resolve: HttpContext
Failed to resolve: HttpContext
Failed to resolve: HttpContext
Failed to resolve: HttpContext
Failed to resolve: HttpContext
Failed to resolve: create
Failed to resolve: path
Failed to resolve: path
Failed to resolve: empty
Failed to resolve: path
Failed to resolve: HttpRuntime
Failed to resolve: create
Failed to resolve: HttpContext
Failed to resolve: HttpContext
Failed to resolve: HttpContext
Failed to resolve: HttpContext
Failed to resolve: HttpContext
Failed to resolve: HttpContext
Failed to resolve: HttpContext
Failed to resolve: HttpContext
Failed to resolve: HttpContext
Failed to resolve: path
Failed to resolve: Http
Failed to resolve: Http.host

Unhandled Exception:
System.Exception: Failed to escape text properly: []
  at RefGen+inCode@356-1.Invoke (System.String message) [0x00001] in <567571a9e56832f5a7450383a9717556>:0
  at Microsoft.FSharp.Core.PrintfImpl+StringPrintfEnv`1[TResult].Finalize () [0x00012] in <5584aad2904cf4daa7450383d2aa8455>:0
  at Microsoft.FSharp.Core.PrintfImpl+Final1@224[TState,TResidue,TResult,A].Invoke (Microsoft.FSharp.Core.FSharpFunc`2[T,TResult] env, A a) [0x00038] in <5584aad2904cf4daa7450383d2aa8455>:0
  at Microsoft.FSharp.Core.OptimizedClosures+Invoke@3253[T2,TResult,T1].Invoke (T2 u) [0x00001] in <5584aad2904cf4daa7450383d2aa8455>:0
[...]

PR #615, opened 2017-04-27
Fix API doc generation

rmunn:
Mismatched backtick in Http.fsi (introduced in commit bb77909829ceeb735a76bc75e8937368533c29dc) was causing API docs to fail.


rmunn:
This should fix #536.

haf:
Excellent! I'm building out the docs directly!

Issue #248, opened 2015-05-15
Validate server-key size on boot


haf:
Repro: Start the web server with a key that is not 32 bytes long.

Details: This happens at runtime, after starting, serving requests:

<blockquote>
System.Exception: internal error on encryption Choice2Of2 (InvalidKeyLength "key should be 32 bytes but was 30
<blockquote>

ademar:
Have you done any research on this? Very weird bug; I can not see how this can happen.

I have seen this one though, it could be related.

https://ci.appveyor.com/project/AdemarGonzalez/suave/build/1.0.505#L296

haf:
Yes, I know the reason of that - that test has a 1/256 chance of failing, see https://github.com/SuaveIO/suave/blob/master/src/Suave.Tests/Utils.fs#L59

We should really zero out enough for it to be very unlikely for a range of zeroes to be there, something like zeroing out 5 bytes; 1/256^5 = 1/1 099 511 627 776

danieljsummers:
This bit me the other day in all my cookie adventures.  :)  Would a single-case DU be a good fix?

Something like...
```fsharp
module ServerKey

  type T = ServerKey of byte[]

  let mkRandom = ServerKey (Crypto.generateKey HttpRuntime.ServerKeyLength)

  let mk (key : byte[]) =
    match key.Length with
    | 32 -> ServerKey key
    | len ->
        [some warning about the length being off, and that we're proceeding with a random one]
        mkRandom

  let value (ServerKey sk) = sk
```

haf:
@danieljsummers 

Thanks for looking into this!

 - I think that the DU case isn't really making it less likely that you generate the wrong key; because we can still have the module named `ServerKey` to generate it
 - Docs on how to call `create` (not `mk`) to generate it
 - I think we should throw like is the case now, otherwise you'll get strange errors of the user session resetting any time you reset a particular process, which makes it even harder to debug in a distributed system setting with many replicas of the web frontend – and you've even set a specific key, so "it should be fine" – better then to crash.

danieljsummers:
OK - great.  I'll work on this once I get my session/state documentation done.  I'm guessing the size-check-on-startup would be [go here, as the first line of `startWebServerAsync`](https://github.com/SuaveIO/suave/blob/f0831e3ed00c611f78a04ac30d3a065403470274/src/Suave/Web.fs#L38).

PR #619, opened 2017-05-29
Validate Server Key (#248) and docs (#368)

danieljsummers:
This creates a `ServerKey` module that will "create" (from a byte array) with validation, or create a key from a base64-encoded string.  On startup, the server runs the configured key through the `create` function to ensure it is well-formed.

This also documents every item in `SuaveConfig`, with examples for some commonly-modified parameters.  There is also verbiage for newer users, pointing them to the three most common "newbie" use cases (changing the port, choosing the home directory, setting a server key).

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -54,7 +54,9 @@ let OPTIONS (x : HttpRequest) = ``method`` "OPTIONS" x
 Server configuration
 --------------------
 
-The first argument to `startWebServer` is a configuration record with the following signature.
+The first argument to `startWebServer` is a configuration record with the
+following signature. (See [below](#changing-the-default-configuration) for tips
+on customizing this.)
 
 {% highlight fsharp %}
 /// The core configuration of suave. See also Suave.Web.defaultConfig which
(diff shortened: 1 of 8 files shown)


src/Suave/Http.fsi
```diff
@@ -331,6 +331,16 @@ module Http =
   /// A server-key is a 256 bit key with high entropy
   type ServerKey = byte []
 
+  /// Utilities to ensure server keys are well-formed
+  [<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
+  module ServerKey =
+    
+    /// Ensure that a server key is the proper length
+    val create : byte [] -> ServerKey
```
ademar: Wouldn't be better called `validate` ?

danieljsummers: Up until I sat down this morning to wrap this up, it literally was.  :)  I can change it back; my thought was that a `byte []` is unknown, but a `ServerKey` is a verified key.  I know that the alias definition means that there's technically no difference between `byte []` and `ServerKey`.  It also seemed that most of the modules have a `create` function.

Let me know if you want me to change it; it's an easy fix.

danieljsummers:
It looks like the AppVeyor build failed for lack of a definition of `ApplicationException` - is there something else I should be using?  (of course it built and tested fine on "my machine"...)

danieljsummers:
I changed `create` to `validate`; I also update the docs and tests for that.  I changed the `ApplicationException` raising to just use `failwithf`.  Finally, I tweaked verbiage in the docs that was inaccurate given the eventual state of my prior PR.

docs/api.md
```diff
@@ -190,3 +192,191 @@ error.
...
+      bindings = [ HttpBinding.createSimple HTTP "127.0.0.1" 8082 ]
+    }
+{% endhighlight %}
+
+As `bindings` is a list, you can also configure Suave to listen on any
+combination of IP addresses and ports.
+
+#### Changing the server's cryptography key
+
+Suave encrypts state stored in (sessions)[sessions.html], and the key used for
```
inosik: This should be `[sessions](sessions.html)`.

danieljsummers: Whoops - nice catch.  The fix has been committed.

PR #620, opened 2017-05-31
Make CORS 'exposeHeaders' to be list of strings

### LINKED ISSUE #576: CORS "Access-Control-Expose-Headers" header is invalid
The code currently use `CORSConfig.exposeHeaders` and place either `"True"` or `"False"` in this header, but it is defined as a list of header names that are allowed in addition to the 6 simple response headers that are always allowed.

https://www.w3.org/TR/cors/#access-control-expose-headers-response-header

> The Access-Control-Expose-Headers header indicates which headers are safe to expose to the API of a CORS API specification. ABNF:

Sample: (From MDN https://developer.mozilla.org/es/docs/Web/HTTP/Headers/Access-Control-Expose-Headers )

```
Access-Control-Expose-Headers: Content-Length, X-Kuma-Revision
```

alex-bogomaz:
Hello, please review this PR. 
It fixes #576 (CORS "Access-Control-Expose-Headers" header is invalid) by making ```exposeHeaders``` to be a predefined list of strings, instead of ```bool```.


ademar:
Thanks!

PR #623, opened 2017-06-12
fix appveyor build

enricosada:
@haf now appveyor is green, what was the issue with .net core sdk?

From the message in chat https://fsharp.slack.com/archives/C1KANN0BV/p1496826486641747 ):
> Hi guys, Suave's build of .Net Core has bitrotted again
Found .NET Core SDK 1.0.4 but require 1.0.1. Downloading and installing in ./tools/coreclr
bash --install-dir "/Users/h/dev/suave/suave/tools/coreclr" --channel "stable" --version "1.0.1
bash: --install-dir: invalid option
Usage:    bash [GNU long option] [option] ...
    bash [GNU long option] [option] script-file ...


haf:
When you follow along the SDK releases on your laptop, you get 1.0.4 installed (of some component) but then the Suave build script stops working with the above error message.

enricosada:
@haf let me check. afaik the build script download and use a specific version in `tools/coreclr`, to exactly avoid this issue. So bump that may work.

Let me check, i'll send a PR enabling .net core on travis

PR #624, opened 2017-06-13
use net core in travis

enricosada:
- [x] build .net core on travis
- [x] fix build script to install and use a specific version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ install:
  - sudo apt-get install -qq apache2-utils
 
 script:
- - bundle && bundle exec rake travis
+ - bundle && bundle exec rake ci
 
 matrix:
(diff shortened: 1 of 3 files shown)


enricosada:
Summary:

- on travis: use ubuntu 14.04 instead of 12.04 because 12 doesnt support .net core 1.0
- fixed the rake task to install .net core on osx/unix (@haf this fix your error locally) when installed version is different than expected
- use the same rake task on both travis and appveyor (renamed from `rake appveyor` to `rake ci`) and so build also .net core + packages, so is easier to check it will continue to work.

enricosada:
@haf ping me is this is not enogh to fix your local issue

haf:
@enricosada This fixed my local issue. Thank you!

PR #633, opened 2017-07-28
Update composing.md

programmer04:
Add information about requiring `open Suave.Utils.Collections`, without this, example can not be compiled due to lack of definition for `^^` operator. Please add this also to documentation on website.


ademar:
Thanks!

Issue #589, opened 2017-03-16
HttpCookie.create with "SameSite" flag as CSRF protection?


Dzoukr:
Hi guys,

I know that support for all browsers is still [not 100%](https://www.chromestatus.com/feature/4672634709082112) (if ever will), but I thing we should think about adding possibility to create HttpCookie with `SameSite` flag (having `Lax` or `Strict` options) to increse security and prevent CSRF attack.

What do you think about this? Is it something worth PR?

haf:
Yes, it's definitely worth PRing.

haf:
Merged.

PR #634, opened 2017-08-01
Support samesite flag of cookie

JacobChang:
this should fix issue #589


JacobChang:
Oh, It seems the continuous integration failed

ademar:
Your code looks OK, the build was already broken.

@haf the build broke after internalizing Yolo module on ba17fa5d04fcf6f471d1a400804bb3901cfc14d6

haf:
+1. Will merge, this is a breaking change, so it should go in the v3 release.

PR #640, opened 2017-08-16
Fix netstandard1.6 DLL having 0.0.0.0 version

### LINKED ISSUE #635: Suave NuGet package has assembly version 0.0.0.0 for netstandard - intended, or mistake?
Yesterday I reported https://github.com/fsprojects/Paket/issues/2576, which @matthid tracked down in https://github.com/Microsoft/visualfsharp/issues/3408 to a version number mismatch in the Suave NuGet package. The `net40` assembly has version number 2.1.1.0, but the `netstandard` assembly has version 0.0.0.0. This caused .NET Core to refuse to load it in my project, because another package (Suave.OAuth 0.9.0-pre2) wanted Suave version 2.0.0.0 or later, and version 0.0.0.0 is less than version 2.0.0.0.

I verified the version numbers by adding Suave to the `paket.dependencies` of a new project, then running `monodis --assembly path/to/Suave.dll | grep Version` to get the assembly version. Here are the results:

```text
$ monodis --assembly packages/Suave/lib/netstandard1.6/Suave.dll | grep Version
Version:       0.0.0.0
$ monodis --assembly packages/Suave/lib/net40/Suave.dll | grep Version
Version:       2.1.1.0
```

Checking `packages/Suave/Suave.nuspec` confirmed that I had indeed grabbed version 2.1.1 of the Suave NuGet package.

Is it intended for the `netstandard1.6` version of the DLL to have a version number of zero? Or is it an omission somewhere in the Suave build process?

rmunn:
This fixes #635 on my machine: the netstandard1.6 DLLs I build now have the correct version number (currently 2.1.1).

Note that in order to get this to build, I had to change the desired dotnet version in the Rakefile from 1.0.1 to 1.0.4: version 1.0.1 of the dotnet CLI tools is apparently no longer available to download from Microsoft's site. I've left that change out of this PR since it's unrelated, but if the build fails, changing `dotnet_version = '1.0.1'` to `dotnet_version = '1.0.4'` in the Rakefile may fix the build.

### DIFF
--- a/src/Suave/Suave.netcore.fsproj
+++ b/src/Suave/Suave.netcore.fsproj
@@ -74,6 +74,7 @@
     <Compile Include="WebSocket.fs" />
     <Compile Include="Owin.fs" />
     <Compile Include="Json.fs" />
+    <Compile Include="AssemblyVersionInfo.fs" Condition="Exists('AssemblyVersionInfo.fs')" />
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="FSharp.NET.Sdk" Version="1.0.*" PrivateAssets="All" />


haf:
This is great! Thanks!

PR #644, opened 2017-08-19
Added an enhanced view engine based on Html.fs and Xml.fs

dustinmoris:
Hi,

I had this on my list of ToDos for a while..., because I extremely like the functional view engine which you have put together I enhanced it quite a bit after coming across a few shortcomings through my own usage. I've tried to improve it by putting it all together in a single "ViewEngine" which can be used for HMTL and XML together.

High level overview of new things:
- Distinguished between raw content and encoded content (there's a `rawText` and `encodedText` functions for that - useful when you want to render inline scripts, etc.)
- Added a `comment` helper function when it is desired to render an actual XML or HTML comment. For example as it is often seen with hidden hiring messages on various websites.
- Extended attributes to be either a key value pair (e.g. `<input type="text" />`) or a single boolean flag (e.g. `<script src=".." async defer>`)
- For this reason you have to set attributes with the helper functions `attr` or `flag` (e.g.: `script [ attr "src" "main.js"; flag "async"; flag "defer" ]`)
- More `renderXYZ` methods which can render either an entire HTML document, or HTML snippets or plain XML, which can be used for things like dynamic SVG images.

Let me know what you guys think and feel free to throw this away and only pick the bits which you like!

### DIFF
--- a/src/Experimental/Suave.Experimental.fsproj
+++ b/src/Experimental/Suave.Experimental.fsproj
@@ -73,6 +73,7 @@
   <Import Project="$(FSharpTargetsPath)" Condition="Exists('$(FSharpTargetsPath)')" />
   <ItemGroup>
     <Compile Include="AssemblyVersionInfo.fs" Condition="Exists('AssemblyVersionInfo.fs')" />
+    <Compile Include="ViewEngine.fs" />
     <Compile Include="Html.fs" />
     <Compile Include="Xml.fs" />
     <Compile Include="Template.fs" />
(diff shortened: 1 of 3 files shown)


ademar:
Great stuff. I am in favor of including your work and abandon the other two modules.

I think we should also move it out of Experimental to its own assembly.

ademar:
:+1: to merge

haf:
Restarting build to see if it's an unrelated error with this PR.

haf:
Yeah, I just read the code for this; it's awesome. But it's missing even the most basic docs so noone will know how to use it.

Could you add a sample to the `examples` folder, perhaps? And if you have time (optional), a very simple Hello World-example in this folder? https://github.com/SuaveIO/suave/tree/master/docs

haf:
LGTM after docs.

haf:
Let's merge it anyway. @dustinmoris Do you think you could write an example in the docs meanwhile?

Issue #616, opened 2017-04-29
userData Persists per Connection Instead of per Request


danieljsummers:
In the discussion of #608, it was discovered that `userData`, which is supposed to have a request lifetime, is preserved across requests when the HTTP keep-alive header is sent.  Even though the connection should be kept open, `userData` should be cleared between requests.

_(When fixing this, be sure to remove the comments from the session documentation referencing the workaround for this issue.)_

JacobChang:
@danieljsummers @haf I'd like to solve this one, do I just need to clear userState at [here](https://github.com/SuaveIO/suave/blob/880f1642f64d385c57f4c60207737f60fcdfea1e/src/Suave/ParsingAndControl.fs#L49) and then remove the comments @danieljsummers mentioned above?

danieljsummers:
That looks like it.  I had tried to figure it out, but couldn't.  It would be good to write a unit test that would fail with the current code (i.e., two requests with keep-alive set, but assume that user data doesn't persist), then implement your fix and see if it passes.

JacobChang:
OK，will submit a PR

PR #647, opened 2017-08-22
userData Persists per Connection Instead of per Request

JacobChang:
This should fix issue #616.


haf:
Thank you @JacobChang for your help and assistance! Keep up the good work! (the code looks good!)

PR #648, opened 2017-08-24
Update DotLiquid

inosik:
References #642. The `Paket.Restore.targets` file probably doesn't need to be updated.


haf:
Thank you @inosik !

PR #658, opened 2017-09-22
Fixes #657

### LINKED ISSUE #657: When serving a compressed file, Suave may miss a later version
When asked to serve a static file that has already been compressed, Suave will compare the modification date on the file with the date of the compressed file, and if it's later, compress it to a new file and serve it. However, if a file is replaced with a later version, but that later version happens to be before the first file was compressed, Suave misses this and doesn't serve the new file. I believe this is a bug, and it needs to compare the date of a file with the previous version of the file, not the timestamp of the compressed file.

I'll submit a PR.

jimfoye:


jimfoye:
This code should build with no problem. I get the same warnings that are emitted by AppVeyor on my development machine, but I don't know why.

haf:
Someone needs to look into the bit-rotting build. Mono's newfangled GC issues don't help...

ademar:
Thanks!

PR #663, opened 2017-10-23
Minor bug fix to error case in generateCookies()

jimfoye:
Missing Choice2Of2 in pattern match (very minor bug).


ademar:
Thanks!

Issue #646, opened 2017-08-22
Bump Expecto in Suave.Testing


seanamos:
Referencing Suave.Testing from a .NET Core/Standard project results in the following warning:

```
C:\xxx.fsproj : warning NU1701: Package 'Expecto 4.2.1' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.0'. This package may not be fully compatible with your project.
C:\xxx.fsproj : warning NU1701: Package 'Mono.Cecil 0.9.6.4' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.0'. This package may not be fully compatible with your project.
C:\xxx.fsproj : warning NU1701: Package 'Suave.Testing 2.2.1' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.0'. This package may not be fully compatible with your project.
```

I have not seen any actual runtime errors as of yet.
There is a workaround to suppress the warning as detailed here: https://github.com/dotnet/standard/issues/439#issuecomment-321608375

However, as far as I know there is currently no way to suppress assembly warnings with paket.

haf:
Expecto v5 supports .Net Core 2. Would you like to do the honours to bump paket.lock?

PR #670, opened 2017-11-28
build "Suave.Testing" for netstandard 1.6

alex-bogomaz:
Hello - please review this PR.
It should fix https://github.com/SuaveIO/suave/issues/646.
We actually need NetStandard build of "Suave.Testing" to get rid of warnings, not increase Expecto version.


ademar:
Thanks!

PR #671, opened 2017-12-20
Run tests under NetCore

alex-bogomaz:
Summary:
- upgraded all ```*.netcore.fsproj``` to 2.0 SDK
- ported ```Suave.LibUv``` to ```netstandard 2.0```,  becase some APIs are missing in ```netstandard 1.6``` (like ```BeginThreadAffinity``` and others)
- run Suave tests under ```netcoreapp 2.0``` (no ```WebSocketSharp``` and  ```Suave.LibUv``` for ```netstandard 1.6```)
- include NetCore tests to CI pipeline
- remove dependency on specific version of Suave for other NetCore NuGet packages

Notes:
- ```Http.fs``` - need to wait for NetCore release. Now using full-framework package.  //cc @seanamos and @ivpadim
- Linux tests are running only with "LibUv TCP engine".  
Reason: some tests (which start and stop server) are failing with "Default TCP engine" - it is not possible on Linux to dispose socket and re-bind - https://github.com/dotnet/corefx/issues/24562. 

We can do something similar as for Kestrel: https://github.com/aspnet/KestrelHttpServer/pull/2111, but this requires building Suave for ```netstandard 2.0```, since there are no ```Handle``` property for ```Socket``` under 1.6.

- Three ```Auth.fs``` tests are excluded because Suave cookie serialization code fails to serialize ```FSharpMap```. Probably that code should not use ```DataContractJsonSerializer``` and use Newtonsoft instead? 

- After converting to 2.0 SDK - there are a lot of build warnings like 
```
Encountered conflict between 'Runtime:..\.nuget\packages\microsoft.win32.registry\4.0.0\runtimes\unix\lib\netstan
dard1.3\Microsoft.Win32.Registry.dll' and 'Runtime:..\olbog\.nuget\packages\microsoft.win32.registry\4.0.0\runtimes\win\lib\nets
tandard1.3\Microsoft.Win32.Registry.dll'.  Could not determine winner due to equal file and assembly versions.
```
but according to this https://github.com/dotnet/standard/issues/505 we may skip those warnings

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: csharp
 dist: trusty
 sudo: required
 
-dotnet: 1.0.4
+dotnet: 2.0.0
 
 mono:
  - 4.8.0
--- a/Rakefile
+++ b/Rakefile
@@ -104,7 +104,7 @@ namespace :dotnetcli do
   end
 
   task :coreclr_binaries => 'tools/coreclr' do
-    dotnet_version = '1.0.1'
+    dotnet_version = '2.0.0'
     dotnet_installed_version = get_installed_dotnet_version
     # check if required version of .net core sdk is already installed, otherwise download and install it
     if dotnet_installed_version == dotnet_version then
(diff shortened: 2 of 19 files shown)


ademar:
This is very very cool, thank you very much :)

ademar:
@alex-bogomaz can we merge this one ?

alex-bogomaz:
Yes - we can merge. 

Netcore tests still depend on net46 "HTTP.fs", but it is OK, since netcore supports referencing full-framework libraries (as long as they work).
We may not wait for netcore release of "HTTP.fs", but rather reference it when it will will be ready.

ademar:
Thanks!

PR #688, opened 2018-02-02
More correct rebinding to same address for NetCore

alex-bogomaz:
This PR is doing same thing for Suave as PR https://github.com/aspnet/KestrelHttpServer/pull/2111
for ```Kestrel.Transport.Sockets```, because following Suave code in ```Tcp.fs``` is not fully correct:
```fsharp
#if NETSTANDARD2_0
listenSocket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.ReuseAddress, 1)		
#endif
```
see also following discussion: https://github.com/dotnet/corefx/issues/24562.

Also two additional things added:
- Add MacOS to CI
- do not build ```libuv``` for Travis - use it from NuGet folder - same as for AppVeyor.

What do you think about this?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,11 @@
 ---
 language: csharp
 
+os:
+  - linux
+  - osx
+
 dist: trusty
-sudo: required
 
 dotnet: 2.0.3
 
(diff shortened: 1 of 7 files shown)


ademar:
Thanks!

PR #712, opened 2018-09-19
Adding support for 303: See Other redirection.

ninjarobot:
HTTP 303 support is useful when you accept a request to create a resource that takes time to create, and, once created, want to communicate the location where information about creating the resource can be found.  A typical flow using HTTP 303 would be as follows:

1. POST to /api/resource - returns `HTTP 202 Accepted`, with a `Location` header indicating `/api/task/1234` can provide details on the progress of creating it.
2. GET `/api/task/1234` - before the task is done, this returns `HTTP 200` with content providing task progress information about the resource being created.  If cancellation is supported, `DELETE /api/task/1234` would be a reasonable way to support it, and the content of the task could return that as a link.
3. GET `/api/task/1234` - once the task is done and the resource is created, this returns `HTTP 303: See Other` which indicates that the resource being created by the task is available now.  A `Location: /api/resource/5678` header redirects the user to the actual resource they tried to create in the first place.
4. GET `/api/resource/5678` - now that the resource is created, this returns `HTTP 200` with the resource's content.

### DIFF
--- a/src/Suave/Combinators.fs
+++ b/src/Suave/Combinators.fs
@@ -178,6 +178,17 @@ module Redirection =
 </html>"
       url HTTP_302.message))
 
+  let see_other url =
+    setHeader "Location" url
+    >=> setHeader "Content-Type" "text/html; charset=utf-8"
+    >=> response HTTP_303 (
+      UTF8.bytes(sprintf "<html>
+  <body>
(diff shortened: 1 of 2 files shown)


ademar:
Thanks!

PR #713, opened 2018-09-26
Fixes multiple problems in last-modified handling

vbfox:
* The header value wasn't in GMT (It was the time on the current server timezone with 'GMT' written after)
* On filesystems with millisecond precision times in fstat a 304 was never returned as the header value is only precise to the second. (All recent linux FS and windows NTFS are in this case)
* Last accessed time was used instead of last modified


ademar:
Thanks!

Issue #727, opened 2019-04-10
Music Store demo, page 370, can't compile with F# 4.6.2


BentTranberg:
I have been working my way through the Music Store demo for two days, and now I got stuck on this weird compiler error when pasting in the source on page 370, chapter **Session type**, from **suave-music-store-en.pdf**

The error is `FS0193 The module/namespace 'Control' from compilation unit 'FSharp.Core' did not contain the val 'ValLinkagePartialKey(OnSuccess)'`

The issue is not spesifically related to Suave, and has been reported here:
https://github.com/Microsoft/visualfsharp/issues/6450

In the issue description, I believe you will find guidance on how this can be resolved in regards to Suave. I am not an expert in these matters, so I could be wrong, but please have a look at it. Anyway I hope to somehow be able to use Suave with the latest FSharp.Core available.

[MusicStore.zip](https://github.com/SuaveIO/suave/files/3065391/MusicStore.zip)

In MusicStore.zip you will find my demo that fails to compile in VS 2019. It does not depend on a database, and it's just compile and run if it compiles. It can be compiled and run if FSharp.Core 4.5.2 is brought in from NuGet, but will otherwise fail with the error mentioned above.

enricosada:
fixing that, i'll send a PR soon

enricosada:
fixed in PR https://github.com/SuaveIO/suave/pull/728

ademar:
@BentTranberg I've pushed a new nuget 2.5.4 containing this fix. Please give it a try and let us know if this does not solve the problem.

PR #728, opened 2019-04-12
fix compile error using FSharp.Core not equal to v4.5

enricosada:
## use FSharp.Core 4.3.4 and constraint that in nuspec

Before this commit the FSharp.Core used to build was the one resolved in paket.lock (`v4.5.2`).
But because `FSharp.Core` is backward compatible, not forward, building against that version may make Suave not compatible when used with older FSharp.Core
An example of bug is https://github.com/SuaveIO/suave/issues/727 

To fix, the minimum version used to build is `v4.3.4` (the first who support netstandard2.0)
In the Suave nuspec, the FSharp.Core is constrained to `>= 4.3.4 && <5`
While the `net461` and `netstandard` target framework may have different FSharp.Core, just one is used to simplify (the v4.3.4 is already old)

## pin .net core sdk, to same version used in travis

Without a `global.json` who pin a specific version, the latest installed in the machine is used and that
may cause issues (new .net core sdk mean new compiler/default fsharp.core/commands/etc)

## remove useless nuget.config

that was an old and now unused tempo feed

## use paket magic mode, upgrade and pin to version 	`5.203.0`

Use paket magic mode (the `paket.exe` is `paket.bootstrapper.exe` renamed) to allow change version just changing the version number in paket.dependencies

Upgrade paket to a recent version (`v5.203.0) and pin the version in paket dependencies to have a deterministic build and get support for bugs fixed using newer .net core sdk

fix https://github.com/SuaveIO/suave/issues/727


src/Suave.DotLiquid/paket.template
```diff
@@ -13,8 +13,6 @@ files
   bin/Release/netstandard2.0/Suave.DotLiquid.pdb => lib/netstandard2.0
 dependencies
   framework: net461
-    FSharp.Core ~> 4
```
enricosada: the `FSharp.Core` is removed from paket.template, so the constraint in `paket.dependencies` will be used
that mean `>= 4.3.4 && <5`

ademar:
@enricosada Thank you very much!

PR #733, opened 2019-07-19
Added method NOT_IMPLEMENTED

isthistechsupport:
The method already exists in Combinators.fs, but not here. This adds it with the respective documentation so it'll (hopefully) show up in Intellisense and not cause a compiler error when calling the method.

### DIFF
--- a/src/Suave/Combinators.fsi
+++ b/src/Suave/Combinators.fsi
@@ -1053,7 +1053,7 @@ module ServerErrors =
   /// The server does not recognize the request method and can not support
   /// it for any resource.
   /// </para></summary>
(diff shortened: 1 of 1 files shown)


ademar:
Hi, thanks for the pull request.

This does not compile as it is. The signature of `NOT_IMPLEMENTED' is incorrect.

```
val NOT_IMPLEMENTED : body:string [] -> WebPart
```

It should be

```
val NOT_IMPLEMENTED : body:string  -> WebPart
```

Can you correct it?

isthistechsupport:
Hi, sorry, I've just corrected it! Thank you for the review

ademar:
Thanks!

Issue #737, opened 2019-11-01
Getting Started Suave on Heroku


dyoshikawa:
Hello.

[The existing heroku tutorial](https://github.com/SuaveIO/suave/blob/master/docs/heroku.md) looks old.

I suggest the following method.

- [Heroku with Docker](https://www.heroku.com/deploy-with-docker)
- ["PublishSingleFile"  in .NET Core 3.0](https://docs.microsoft.com/en-us/dotnet/core/whats-new/dotnet-core-3-0#compiledeploy)

I made the sample.

https://github.com/dyoshikawa/heroku-suave-getting-started

This is a way for developers to develop on Heroku in a simple, easy and modern environment.

I want to add this to the documentation.
I will contribute if you feel good.

ademar:
Hi @dyoshikawa, your contribution would be welcome 👍

dyoshikawa:
@ademar Thanks for reading.
I will prepare immediately.

PR #738, opened 2019-11-08
Add Heroku with Docker instructions

dyoshikawa:
This is a proposal to introduce the .NET Core 3.0 single file generation and Heroku deployment method using Docker.

https://github.com/SuaveIO/suave/issues/737


ademar:
Thanks!

PR #747, opened 2020-12-04
Fixing sequence map bug for getHeaders

afzhaq-msft:
Fixing sequence map bug for getHeaders which only takes labels instead of values


ademar:
Thanks!

Issue #759, opened 2022-01-21
Issue with parsing boundary from multipart/form-data Content-Type header causes a crash


eriadam:
Hello!

We found that in case the Content-Type header contains a `charset` parameter next to the boundary, the parsing logic is faulty, which leads to an invalid boundary and a crash.

```
Content-Type: "multipart/form-data; charset=utf-8; boundary=__X_PAW_BOUNDARY__"
```

results in

```
 --utf-8; boundary=__X_PAW_BOUNDARY__
```

as the logic in the `ConnectionFacade.fs` looks for the first `=`.

```fsharp
let boundary =
    "--"
    + (ce
       |> String.substring (ce.IndexOf('=') + 1)
       |> String.trimStart
       |> String.trimc '"')
```

<details>
  <summary>Click to expand the stack trace</summary>

```
[12:23:24 WRN] TCP request processing failed
System.Exception: Invalid multipart format: expected boundary ‘--utf-8; boundary=__X_PAW_BOUNDARY__’ got ‘--__X_PAW_BOUNDARY__’
   at Microsoft.FSharp.Core.PrintfModule.PrintFormatToStringThenFail@1439.Invoke(String message) in D:\a\_work\1\s\src\fsharp\FSharp.Core\printf.fs:line 1439
   at <StartupCode$Suave>.$ConnectionFacade.clo@413-21.Invoke(String _arg18)
   at Microsoft.FSharp.Control.AsyncPrimitives.CallThenInvokeNoHijackCheck[a,b](AsyncActivation`1 ctxt, b result1, FSharpFunc`2 userCode) in D:\a\_work\1\s\src\fsharp\FSharp.Core\async.fs:line 464
   at <StartupCode$Suave>.$SocketMonad.Return@9-1.Invoke(AsyncActivation`1 ctxt)
   at <StartupCode$Suave>.$SocketMonad.Return@9-1.Invoke(AsyncActivation`1 ctxt)
   at <StartupCode$Suave>.$SocketMonad.Return@9-1.Invoke(AsyncActivation`1 ctxt)
   at Suave.Sockets.SocketOpModule.orInputError@46-1.Invoke(AsyncActivation`1 ctxt)
   at <StartupCode$Suave>.$SocketMonad.Return@9-1.Invoke(AsyncActivation`1 ctxt)
   at <StartupCode$Suave>.$SocketMonad.Return@9-1.Invoke(AsyncActivation`1 ctxt)
[...]

PR #760, opened 2022-01-21
Using RegEx for parsing the boundary from the ContentType header.

eriadam:
As detailed in the issue #759, the current parsing of the boundary does not respect the `charset` property.

The PR proposes a fix to use a regular expression to parse the boundary from the ContentType header value based on [RFC2046](https://www.rfc-editor.org/rfc/rfc2046#section-5.1.1).  It allows alphanumeric characters, punctuations and spaces (except at the end). Quotation marks seem to be optional as well.

```fsharp
let parseBoundary contentType =
  let pattern = "boundary=\"?([a-zA-Z0-9'\(\)+_,-.\/:=? ]*)(?<! )\"?"
  Regex.Match(contentType, pattern).Groups.[1].Value
```

```fsharp
| Choice1Of2 ce when String.startsWith "multipart/form-data" ce ->
  let boundary = "--" + parseBoundary ce
```

Changed for `"multipart/form-data"` and `"multipart/mixed"` in the code.

---

Input:

```
Content-Type = "multipart/form-data; charset=utf-8; boundary=__X_PAW_BOUNDARY__"
```

Original result:
```
 --utf-8; boundary=__X_PAW_BOUNDARY__
```

New results:
```
'multipart/form-data; charset=utf-8; boundary=__X_PAW_BOUNDARY__' => __X_PAW_BOUNDARY__
'multipart/form-data; boundary=__X_PAW_BOUNDARY__' => __X_PAW_BOUNDARY__
'multipart/form-data; boundary=__X_PAW_BOUNDARY__ charset=utf-8;' => __X_PAW_BOUNDARY__
'multipart/form-data;' => (null)
```

### DIFF
--- a/src/Suave.Tests/Parsing.fs
+++ b/src/Suave.Tests/Parsing.fs
@@ -14,6 +14,7 @@ open Suave.Successful
 open Suave.Utils
 open Suave.Tests.TestUtilities
 open Suave.Testing
+open Suave.Utils.Parsing
 
 [<Tests>]
 let parseQuery =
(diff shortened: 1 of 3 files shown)


ademar:
Hi @eriadam 

Your PR as it is now it breaks a bunch of tests. The problem is the pattern you are using only selects word characters while boundaries can include all sorts of non white space characters. I think the correct regex pattern would be ` "boundary=(\S*)"` instead.

Also for completeness it would be nice if you would include a test case.

Thanks!

eriadam:
Hi @ademar ,
I updated the solution, please have a look if it is OK like this.
Cheers

ademar:
It looks great, thanks very much.  I'll push a new nuget version with these changes tomorrow.

PR #777, opened 2023-08-13
Various improvements to documentation

valhallasw:
My main goal was to improve the docs of the query parameter methods, but I had to solve some build issues along the way so updated the docs on those steps as well.


ademar:
Thanks!

PR #828, opened 2026-05-30
Fix CPU spin on half-closed WebSocket / silent body truncation

ademar:
readPostData aborts silently on EOF or cancellation, leaving the caller with a zero-filled buffer and no indication that the read was short.

In WebSocket.readBytes / readBytesIntoByteSegment this caused the frame parser to see opcode=0 length=0 (a zero-length Continuation frame), return immediately, and loop forever - pegging a CPU core after the peer half-closed the connection.

In ConnectionFacade.getRawPostData the same bug silently truncated HTTP/1 request bodies, returning Ok with a partially-zeroed buffer instead of surfacing the broken connection.

The HTTP/2 reader (Http2.readBytes) already guards against this with an 'if offset = n' check; apply the same fix to the other three call sites.

Added a regression test in `src/Suave.Tests/WebSocket.fs` that opens a raw TCP socket, performs a valid WebSocket upgrade handshake, half-closes the client send side, and asserts the handler exits within a deadline.

### DIFF
--- a/src/Suave.Tests/WebSocket.fs
+++ b/src/Suave.Tests/WebSocket.fs
@@ -3,6 +3,8 @@ module Suave.Tests.WebSocket
 open Expecto
 
 open System
+open System.IO
+open System.Net.Sockets
 open Websocket.Client
 open System.Text
 open System.Threading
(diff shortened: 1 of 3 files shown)


ademar:
@copilot Can you produce a test that opens a TCP socket, sends a valid WebSocket upgrade, then half-closes — and asserts the handler exits within a deadline.

PR #844, opened 2026-08-29
Fix static file fallback for missing resources

krymtkts:
`ServeResource.resource` calculated an ETag before checking whether a resource existed.
For missing files, `Files.browseHome` returned HTTP 500 instead of falling through to the next `WebPart`.
This change checks for existence first and adds a regression test for fallback behavior.


ademar:
Thank you!

PR #850, opened 2026-09-07
Add hideStartupMessage config flag to silence the listener banner

### LINKED ISSUE #849: Allow suppressing the "Smooth! Suave ... listener started" console line
`Tcp.fs` writes the startup line straight to the console:

```fsharp
Console.WriteLine($"Smooth! Suave v{Globals.SuaveVersion} listener started in {startedListeningMilliseconds} ms with binding {ipAddress}:{port}")
```

(https://github.com/SuaveIO/suave/blob/master/src/Suave/Tcp.fs, in `runServer`, and the `({effective} acceptors)` variant next to it.)

Since Suave 3 dropped `Suave.Logging`, there is no configuration path that routes or silences this line: `SuaveConfig` has `hideHeader`, but that only removes the HTTP `Server` response header. Applications that embed Suave as a dev server and manage their own console output (in our case `fsdocs watch`, which has `--verbosity quiet`) cannot keep their output clean without redirecting `Console.Out` for the whole process.

Would you accept one of the following?

1. A `SuaveConfig` flag (for example `quiet: bool` or `hideStartupMessage: bool`) that skips the `Console.WriteLine` calls in `runServer`, or
2. Routing the message through a configurable callback, e.g. `SuaveConfig.onListening: (IPAddress * int -> unit)` with the current `Console.WriteLine` as the default.

nojaf:
Apps that embed Suave as a dev server (e.g. fsdocs watch --verbosity quiet) had no way to keep console output clean short of redirecting Console.Out for the whole process, since hideHeader only affects the HTTP Server response header. SuaveConfig.hideStartupMessage (and the matching HttpRuntime field) now lets callers opt out of the "Smooth! Suave ... listener started" line printed in Tcp.fs. Defaults to false, so existing behavior is unchanged. 

Fixes #849.

### DIFF
--- a/examples/Example/Program.fs
+++ b/examples/Example/Program.fs
@@ -181,6 +181,7 @@ let main argv =
       compressedFilesFolder = None
       cookieSerialiser      = new BinaryFormatterSerialiser()
       hideHeader            = false
+      hideStartupMessage    = false
       maxContentLength      = 1000000
       healthCheckEnabled    = true
       healthCheckIntervalMs = 30000
--- a/src/Suave/Http.fs
+++ b/src/Suave/Http.fs
@@ -409,6 +409,9 @@ module Http =
       matchedBinding    : HttpBinding
       cookieSerialiser  : CookieSerialiser
       hideHeader        : bool
+      /// Make this true to skip the "Smooth! Suave ... listener started" console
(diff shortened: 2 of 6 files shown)


ademar:
Thank you!

